PR #2, opened 2015-08-18
AddFetchTimeout discards curl options if 'timeout' is not defined.

deweerdt:
Editing the options passed to curl via 'my @URL_FETCHER = ("curl", "-s");' (in particular to add a -k to ignore self signed certs) fails for some invocations of curl. In FetchDynamicProfile, 'my @fetcher = AddFetchTimeout($fetch_timeout, @URL_FETCHER);' ends up being just 'curl' if timeout is not defined.

This happens because AddFetchTimeout doesn't retrieve all the arguments from the caller.

### DIFF
--- a/src/pprof
+++ b/src/pprof
@@ -3410,7 +3410,7 @@ sub ResolveRedirectionForCurl {
 # Add a timeout flat to URL_FETCHER.  Returns a new list.
 sub AddFetchTimeout {
   my $timeout = shift;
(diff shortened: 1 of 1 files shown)


deweerdt:
I ended up editing the comment inline, let me know if that clears things up a bit.

alk:
Merged out-of-band after editing commit message. Thanks a lot for this commit.

PR #714, opened 2015-09-21
685 Realloc is not supposed to zero all allocated memory

sgajera:


alk:
Big thanks for your effort.

But this patch is changing windows-only recalloc (note 'C' in the middle) function which actually needs to zero new bits of memory it if allocation size grows. The issue we have is that it seemingly zeroes all memory.

gshirishfree:
Hi Alk,

What do you think for below solution ?

extern "C" void* _recalloc(void* p, size_t n, size_t size) {	
   void* result = calloc(1, n * size);
   memcpy(result, p, tc_malloc_size(p));
   free(p);
   return result;	
 }

alk:
I think it'll work.

We could try to do even better by trying to avoid always reallocating object, but given that we don't really know original allocation size, I don't think it'll work.

We can optimize shrinking case which doesn't need to memset anything if you want.

alk:
Sorry for late reply. Please rebase this into single patch.

p == NULL case is best moved to the top. Also no need to set calloc result to variable and then return it. Just straight return is best option.

Also I think we can do slightly better by using realloc and only zeroing part after prev_size until the end.

alk:
This has been superseded by https://github.com/gperftools/gperftools/pull/994 which was just merged. Thanks for reporting the problem.

Issue #699, opened 2015-08-23
It doesn't work on s390x architecture 


alk:
Originally reported on Google Code with ID 696
```
What steps will reproduce the problem?
1. execute make on a s390x linux server
2. receive an error

What is the expected output? What do you see instead?
:/tmp/gperftools-2.4# make
/bin/bash ./libtool  --tag=CXX   --mode=compile g++ -DHAVE_CONFIG_H -I. -I./src  -I./src
  -DNO_TCMALLOC_SAMPLES -pthread -DNDEBUG -Wall -Wwrite-strings -Woverloaded-virtual
-Wno-sign-compare -fno-builtin-malloc -fno-builtin-free -fno-builtin-realloc -fno-builtin-calloc
-fno-builtin-cfree -fno-builtin-memalign -fno-builtin-posix_memalign -fno-builtin-valloc
-fno-builtin-pvalloc  -Wno-unused-result   -g -O2 -MT src/libtcmalloc_minimal_la-tcmalloc.lo
-MD -MP -MF src/.deps/libtcmalloc_minimal_la-tcmalloc.Tpo -c -o src/libtcmalloc_minimal_la-tcmalloc.lo
`test -f 'src/tcmalloc.cc' || echo './'`src/tcmalloc.cc
libtool: compile:  g++ -DHAVE_CONFIG_H -I. -I./src -I./src -DNO_TCMALLOC_SAMPLES -pthread
-DNDEBUG -Wall -Wwrite-strings -Woverloaded-virtual -Wno-sign-compare -fno-builtin-malloc
-fno-builtin-free -fno-builtin-realloc -fno-builtin-calloc -fno-builtin-cfree -fno-builtin-memalign
-fno-builtin-posix_memalign -fno-builtin-valloc -fno-builtin-pvalloc -Wno-unused-result
-g -O2 -MT src/libtcmalloc_minimal_la-tcmalloc.lo -MD -MP -MF src/.deps/libtcmalloc_minimal_la-tcmalloc.Tpo
-c src/tcmalloc.cc  -fPIC -DPIC -o src/.libs/libtcmalloc_minimal_la-tcmalloc.o
In file included from src/tcmalloc.cc:117:0:
src/base/basictypes.h:361:5: error: #error Could not determine cache line length -
unknown architecture
 #   error Could not determine cache line length - unknown architecture
     ^
Makefile:4417: recipe for target 'src/libtcmalloc_minimal_la-tcmalloc.lo' failed
[...]

PR #722, opened 2015-09-30
Some fixes and removal of unportable cycleclock

alk:
Hi. This is another sequences of patches for review.

It includes:

* fix for segfault in profiling handler

* removal of tcmalloc_minimal hard dependency on libpthread (it is using weak symbols as before to detect libpthread presence)

* cleanup two users of cycleclock (spinlock and sysinfo)

* removal of cycleclock (which is very unportable; see issue #699)


alk:
Thanks. Merged with corrections.

PR #725, opened 2015-10-01
Ensure ThreadCache objects are CACHELINE_ALIGNED.

refnum:
The ThreadCache class comments that it is 'critical for performance' that it's correctly aligned, which it enforces with the CACHELINE_ALIGNED attribute.

ThreadCache objects are allocated by ThreadCache::NewHeap. This uses threadcache_allocator to obtain memory, where threadcache_allocator is a PageHeapAllocator.

PageHeapAllocator::New uses MetaDataAlloc to allocate memory, which returns blocks from space allocated with TCMalloc_SystemAlloc.

TCMalloc_SystemAlloc uses a MemoryAligner union to ensure alignment is always CACHELINE_ALIGNED in size.

MetaDataAlloc uses a void* for alignment so it may return pointers with a smaller alignment.

This causes ThreadCache objects to be void*-aligned rather than CACHELINE_ALIGNED.

This was detected by Clang's Undefined Behavior Sanitizer on x86_64.

This breaks in ThreadCache::NewHeap because the pointer assigned to the ThreadCache instance does not match the expected CACHELINE_ALIGNED alignment of the ThreadCache class.

### DIFF
--- a/src/base/basictypes.h
+++ b/src/base/basictypes.h
@@ -364,6 +364,12 @@ class AssignAttributeStartEnd {
 # define CACHELINE_ALIGNED
 #endif  // defined(HAVE___ATTRIBUTE__) && (__i386__ || __x86_64__)
 
+// Structure for discovering alignment
+union MemoryAligner {
+  void*  p;
+  double d;
+  size_t s;
+} CACHELINE_ALIGNED;
 
 // The following enum should be used only as a constructor argument to indicate
 // that the variable has static storage class, and that the constructor should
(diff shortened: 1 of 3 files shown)


alk:
Massive thanks for spotting and reporting it.

You're right. I'll think about your patch a bit more. Only thing that I'm not sure about (yet) is MemoryAligner in common.h.

refnum:
Sure -I took that approach just because MemoryAligner already existed so it seemed plausible to reuse it.

Another approach might be for the per-architecture conditionalisation to define a CACHELINE_SIZE constant. The CACHELINE_ALIGNED attribute then specifies the attribute in terms of that and the other alignment constants could just assign it directly rather than sizeof()-ing anything.

As ThreadCache is the only MetaDataAlloc-client that has this alignment requirement then another variation might be for it to pass the alignment in. The current approach makes for a simpler patch but as the other instances only need 4/8-byte alignment it might be a bit excessive to make them always use 64.

Let me know if you want a revised approach and I'll update.

alk:
Merged. Thanks.

PR #731, opened 2015-10-15
Fixups for MIPS and callgrind format

bo-on-software:
Fixed typo in passing arguments to _syscall6 - it is correctly stated in comment that old MIPS ABI use only 4 registers for argument passing, but the code was trying to pass arg5 and arg6 as registers too.

Also, fixed printf misuse in pprof - the printed string is passed as format. It can cause troubles if string contains "%".


alk:
Applied. Thanks a lot.

PR #732, opened 2015-10-20
avoid double patching when EnumProcessModules fail

qq325895369:
when i use tcmalloc,I found a fatal bug in function PatchAllModules,in my own project,link libtcmalloc,encountered a  dead loop problem, so I generated a full dmp file,after my analyzing,then i located here:
once EnumProcessModules fail,when next  time to execute PatchAllModules,if this api call successfully,will lead to double patching,then lead to dead loop finally！（perftools_free<1> jmp to perftools_free<0>,but perftools_free<0> jmp back to perftools_free<1>）

this bug often appears!!!   and you can reappear it by making the EnumProcessModules call fail manually
if not,you can get the dmp file here: http://pan.baidu.com/s/1kTAhEjd

sorry,my english is not good,can you get it?

### DIFF
--- a/src/windows/patch_functions.cc
+++ b/src/windows/patch_functions.cc
@@ -530,8 +530,10 @@ bool LibcInfo::PopulateWindowsFn(const ModuleEntryCopy& module_entry) {
 
   bool found_non_null = false;
   for (int i = 0; i < kNumFunctions; i++) {
-    if (windows_fn_[i])
+    if (windows_fn_[i]){
       found_non_null = true;
+	  break;
+	}
   }
   if (!found_non_null)
     return false;
@@ -693,7 +695,9 @@ bool PatchAllModules() {
   HMODULE hModules[kMaxModules];  // max # of modules we support in one process
   if (!::EnumProcessModules(hCurrentProcess, hModules, sizeof(hModules),
                             &num_modules)) {
-    num_modules = 0;
+	//qq325895369:must return to avoid double patching,otherwise,
+	//memory functions of tcmalloc self will alse be patched,then lead to dead loop!
+	return false;
   }
   // EnumProcessModules actually set the bytes written into hModules,
   // so we need to divide to make num_modules actually be a module-count.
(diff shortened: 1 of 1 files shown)


qq325895369:
when i use tcmalloc,I found a fatal bug in function PatchAllModules,in my own project,link libtcmalloc,encountered a  dead loop problem, so I generated a full dmp file,after my analyzing,then i located here:
once EnumProcessModules fail,when next  time to execute PatchAllModules,if this api call successfully,will lead to double patching,then lead to dead loop finally！（perftools_free<1> jmp to perftools_free<0>,but perftools_free<0> jmp back to perftools_free<1>）

this bug often appears!!!   and you can reappear it by making the EnumProcessModules call fail manually
if not,you can get the dmp file here: http://pan.baidu.com/s/1kTAhEjd

sorry,my english is not good,can you get it?

qq325895369:
refer to issue #733 @alk @wizeman

src/windows/patch_functions.cc
```diff
@@ -765,13 +769,19 @@ bool PatchAllModules() {
       }
     }
 
+	//qq325895369:on windows os,the for loop above has already pathched all runtime lib, 
+	//delete following code fragment to avoid double patching,otherwise,
+	//memory functions of tcmalloc self will alse be patched,then lead to dead loop!
+#ifndef WIN32
```
alk: Why ifndef? Aren't we always WIN32 here?

src/windows/patch_functions.cc
```diff
@@ -530,8 +530,10 @@ bool LibcInfo::PopulateWindowsFn(const ModuleEntryCopy& module_entry) {
 
   bool found_non_null = false;
   for (int i = 0; i < kNumFunctions; i++) {
-    if (windows_fn_[i])
+    if (windows_fn_[i]){
       found_non_null = true;
+	  break;
```
alk: please fix indentation.

src/windows/patch_functions.cc
qq325895369: yes, "#ifndef" is redundant, useless

alk:
I apologize for any possible misunderstanding here. I am still struggling to understand what you're trying to fix here.

Thing is, whenever enum modules fails, previous code would assume no modules and then unpatch them, other than main module. So no double patching.

Also main module will never be patched twice, given logic.

So I don't understand at all what your fix is doing.

I do find locking in this code is quite questionable w.r.t. concurrent calls to PatchAllModules, especially if we assume enum modules could fail. In those cases I can see how some pieces other than main module, could be patched twice potentially.

alk:
Closing as stale. Please reopen if/when you respond. I am sorry we could not reach understanding on this (possible but in gperftools).

PR #734, opened 2015-10-26
Add generic.total_physical_bytes property to Mallocextension

ssiddhartha:
The actual physical memory usage of tcmalloc cannot be obtained by
GetNumericProperty. This accounts for the current_allocated_bytes,
fragmentation and malloc metadata, and excludes the unmapped memory
regions. This helps the user to understand how much memory is actually
being used for the allocations that were made.

### DIFF
--- a/src/gperftools/malloc_extension.h
+++ b/src/gperftools/malloc_extension.h
@@ -160,6 +160,14 @@ class PERFTOOLS_DLL_DECL MallocExtension {
   //            freed memory regions
   //      This property is not writable.
   //
+  //  "generic.total_physical_bytes"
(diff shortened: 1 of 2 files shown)


alk:
Thanks. But this fails to compile since we don't have metadata_unmapped_bytes metric in TCMallocStats.

src/tcmalloc.cc:665:24: error: 'struct TCMallocStats' has no member named 'metadata_unmapped_bytes'
                - stats.metadata_unmapped_bytes
                        ^

ssiddhartha:
Many thanks for trying the patch. I think this patch is useful only for chromium modified tcmalloc. I just realized. Sorry.

PR #743, opened 2015-11-20
Fix tmpdir usage in heap-profiler_unittest.sh

### LINKED ISSUE #742: Tests fail when run in parallel
Multiple tests are using the same TMPDIR, so running them in parallel with something like

    make check-TESTS -j8

will pretty reproducibly fail with an error like

```
FAIL: heap-profiler_debug_unittest.sh
=====================================

rm: cannot remove '/tmp/heap_profile_info': Directory not empty
mkdir: cannot create directory '/tmp/heap_profile_info': File exists
```

This affects e.g. Debian package builds, where the tests are run as part of the build process and people want to use parallel compilation for performance reason.

Is there a reason not to use a dedicated TMPDIR for each test?

jrosenboom:
Using a single fixed directory would break when tests were being run in
parallel with "make -jN".

Also, the cleanup at the end of the test didn't work because it referred
to the wrong variable.

Fixes #742.

### DIFF
--- a/src/tests/heap-profiler_unittest.sh
+++ b/src/tests/heap-profiler_unittest.sh
@@ -54,13 +54,10 @@ fi
 
 HEAP_PROFILER="${1:-$BINDIR/heap-profiler_unittest}"
 PPROF="${2:-$PPROF_PATH}"
-TEST_TMPDIR=/tmp/heap_profile_info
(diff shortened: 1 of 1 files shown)


alk:
Applied. Thanks a lot.

PR #746, opened 2015-11-22
Central cache stats

sgajera:


alk:
Here is my thinking regarding use of std::vector.

My concern is that ABI behind std::vector may change from compiler to
compiler or from compiler version to compiler version. On GNU systems it
seems unlikely that vector's representation might change. But on Windows it
seems to be common for newer versions of MSVC to have different C++ ABI.

So my concern is if we rely on std::vector in our API and ABI, and somebody
links to tcmalloc which is built with "wrong" ABI for std::vector, then
things will break.

Another thing is that it is generally good idea to always have matching
malloc_extension_c API for all things that we support in C++ API. So that
is perhaps good enough reason not to rely on std::vector.

gshirishfree:
uploded new patch for fixing comment. I'll think of any better way compare to vector.

gshirishfree:
Hi,

Can you take a look at this patch set ? Some how I am not able to merge this patch in one

alk:
I have rebased your patch and re-uploaded it here: https://github.com/alk/gperftools/commits/wip-central-cache-stats-patch

The patch looks ok. But since it is API that we'll have to support for long time I'll need to think about it a bit more. Ideal API is as generic and as implementation-independent as possible.

What is your use case for this new API? Where and how you're planning to use it ?

gshirishfree:
People allocate memory and forget to free memory, if such thing in there in data path that will eventually lead to memory leak. So, to catch this thing like early in stage.

I have seen in my application numerous memory leak, basically my system run out-of-memory after 15 days, the reason is there in leak in one of the central cache class. So, if there is API to get current information of class I can analyze the class information early in stage and catch there is leak in specific class.
I think this is general use case.

alk:
I see. But how exactly you're going to use those _free list_ stats in order to deal with leaks ?

Also why GetStats isn't enough?

gshirishfree:
GetStats gave information about free memory, not inuse memory. Also GetStats give additional information, which I am not interested. This API give information about inuse, free, and no additional text parsing just simple array iteration.

gshirishfree:
Once I know the leak is from this class, from my application I will see who all are allocating near to this class size using malloc_size or tcmalloc_size and find out who is leaking.

We have our own override global new and delete there I'll track this class size, who are allocating and not freeing. OR There is generic API AddNewHook where people can track this specific class size.

You can argue pprof heap profiler will do this job for you, but it's very slow and it dump lot's of data because it track every class size.

Please let me know if it's not clear.

alk:
On Sun, Dec 13, 2015 at 9:47 AM, gshirishfree <notifications@github.com>
wrote:

> Once I know the leak is from this class, from my application I will see
> who all are allocating near to this class size using malloc_size or
> tcmalloc_size and find out who is leaking.
>
> We have our own override global new and delete there I'll track this class
> size, who are allocating and not freeing. OR There is generic API
> AddNewHook where people can track this specific class size.
>
> You can argue pprof heap profiler will do this job for you, but it's very
> slow and it dump lot's of data because it track every class size.
>
> Please let me know if it's not clear.
>

There is actually much better way of dealing with leaks. Take a look at
heap profiler or see lighter-weight MallocExtension::GetHeapSample or
GetHeapGrowthStacks. All this approaches enable you to quickly hunt down
leaks by comparing profile samples.

gshirishfree:
I'll take a look at the API, but this is after we found there is leak.

To know there is leak we need this new API. Along with this it will give information about how memory is getting allocated, which bucket is used more and if they merge there usage in bucket etc etc.

alk:
On Mon, Dec 14, 2015 at 10:06 AM, gshirishfree <notifications@github.com>
wrote:

> I'll take a look at the API, but this is after we found there is leak.
>
> To know there is leak we need this new API. Along with this it will give
> information about how memory is getting allocated, which bucket is used
> more and if they merge there usage in bucket etc etc.
>

All of that info is available via heap sampling. Which can be always on. At
google it is always on for all services as far as I am aware for a very
small overhead.

gshirishfree:
I still believe this API will add more useful information for client, Unnecessary enabling heap sampling be default is not good option at-least for my high performance application, debug information should be enable when only needed.

alk:
On Mon, Dec 14, 2015 at 10:59 AM, gshirishfree <notifications@github.com>
wrote:

> I still believe this API will add more useful information for client,
> Unnecessary enabling heap sampling be default is not good option at-least
> for my high performance application, debug information should be enable
> when only needed.
>

Let me try to explain my thinking.

There is already existing legacy of malloc stats api's exposing too much of
implementation details which make no sense in other implementations. Take a
look at man mallinfo (and compare GNU version and freebsd version) as
example.

It's hard or maybe even impossible to expose useful info without tying it
too much to implementation details. In your case implementation detail is
size classes (and your api doesn't even expose info about large allocations
that are greater than largest size class).

So unless really necessary, I am reluctant to add very non-extensible
(there is no way to add or remove any fields in the struct) and
implementation-specific (what if we drop size classes or e.g. start
managing middle-sized objects differently?) API such us this.

And therefore I am asking why you need this. So far I don't see very strong
case for really needing such API. Specifically:

a) you can see "if there is leak" by using existing stats (if heap grows
that's worth checking for leak)

b) heap sampling is really effective and very low-overhead solution. It is
cheap enough so that it is kept enabled even for most performance and cost
sensitive services at Google.


BTW I was wrong about vectors. Since it is already used by one of methods
in MallocExtension it is okay to keep using it.

PR #748, opened 2015-12-02
fix compilation error in spinlock

cyshi:
Fix compilation error with `__builtin_ia32_pause()`  in gcc-3.4.5


alk:
Thanks a lot. Indeed I've forgotten to check __builting_ia32_pause support for older compilers.

PR #750, opened 2015-12-02
Add support for obtaining cache size of the current thread and softer…

acmorrow:
… idling

### DIFF
--- a/src/gperftools/malloc_extension.h
+++ b/src/gperftools/malloc_extension.h
@@ -395,6 +395,15 @@ class PERFTOOLS_DLL_DECL MallocExtension {
   // Like ReadStackTraces(), but returns stack traces that caused growth
   // in the address space size.
   virtual void** ReadHeapGrowthStackTraces();
+
+  // Returns the size in bytes of the calling threads cache.
+  virtual size_t GetThreadCacheSize();
+
+  // Like MarkThreadIdle, but does not destroy the internal data
+  // structures of the thread cache. When the thread resumes, it wil
+  // have an empty cache but will not need to pay to reconstruct the
+  // cache data structures.
+  virtual void MarkThreadTemporarilyIdle();
 };
 
 namespace base {
--- a/src/gperftools/malloc_extension_c.h
+++ b/src/gperftools/malloc_extension_c.h
@@ -79,6 +79,8 @@ PERFTOOLS_DLL_DECL void MallocExtension_ReleaseToSystem(size_t num_bytes);
 PERFTOOLS_DLL_DECL void MallocExtension_ReleaseFreeMemory(void);
 PERFTOOLS_DLL_DECL size_t MallocExtension_GetEstimatedAllocatedSize(size_t size);
 PERFTOOLS_DLL_DECL size_t MallocExtension_GetAllocatedSize(const void* p);
+PERFTOOLS_DLL_DECL size_t MallocExtension_GetThreadCacheSize(void);
+PERFTOOLS_DLL_DECL void MallocExtension_MarkThreadTemporarilyIdle(void);
 
 /*
  * NOTE: These enum values MUST be kept in sync with the version in
--- a/src/malloc_extension.cc
+++ b/src/malloc_extension.cc
@@ -193,6 +193,14 @@ void MallocExtension::GetFreeListSizes(
   v->clear();
 }
 
+size_t MallocExtension::GetThreadCacheSize() {
+  return 0;
+}
+
+void MallocExtension::MarkThreadTemporarilyIdle() {
+  // Default implementation does nothing
+}
+
 // The current malloc extension object.
 
 static MallocExtension* current_instance;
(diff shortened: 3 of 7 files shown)


acmorrow:
Please consider this a work-in-progress, as we are still evaluating whether or not this is exactly what we want for MongoDB, which is currently using some (slightly different) customizations to its bundled copy of gperftools.

This patch does two things:

- Allows the current thread to obtain the size of its thread cache
- Allows the current thread to drop memory from its cache, without destroying the cache data structures itself.

The idea is that with these two changes, we can, before a thread goes idle, check if it is caching more than some high watermark of memory, and if so, release that memory *without* destroying the threads cache data structures. This way, when the thread does resume, it doesn't pay as heavy a penalty on its first allocation.

We are still evaluating whether this works as we hope, but we have some preliminary evidence that it does, so I'd like to get the review started sooner rather than later. In particular, I'd hope that we could get this change in for the (presumably) upcoming gperftools 2.6.

Note that this currently doesn't have any tests. If you have suggestions on how to write good tests for this I'm happy to oblige.

alk:
Looks good. Testing-wise we should at least make sure that new code is exercised. It should be not too hard to do. You should try testing desired outcome of MarkThreadTemporarilyIdle.

Here is alternative idea. When thread becomes idle it can get it's heap into 'heap free list'. And when it stops being idle it can grab it from free list. We can make those free lists per cpu if needed. With this design you should get:

a) sharing of heaps and thus less memory waste than otherwise

b) very quick "become idle/unbecome idle" operations. Much faster than TemporarilyIdle and friends.

What you think?

acmorrow:
@alk - Per-CPU has been on my mind as well, and I have some time dedicated to exploring it next week. I'll add your idea to the set of experiments I want to run.

Just to make sure we are on the same page, what you are suggesting is that there be:
- A free heap list (scoped either globally or per-cpu)
- When a thread goes idle, it just transfers its entire heap onto a/the heap free list
- When that thread next performs a (de)allocation, it will see that it has no heap. Before constructing one anew, it first checks the free list and if it finds one there, it uses it.

That does sound pretty interesting. How would stealing work? I presume that if a thread needed to steal it would do so from the heaps on the free list first?

Can you give me any idea how much time we have to work towards up solution before gperftools 2.6 happens?

alk:
Your list mostly matches what I proposed. The only difference is that I was thinking about some explicit call when thread goes out of idle. It might end up insignificant, but my thinking is that we want randomly created new threads not to have warmed up heap. Ideally we'll only use heaps from free list for your worker threads. Thus separate (and cheap) activation call. But as I noted above it'll probably work too without explicit activation.

Regarding my time in new couple months. I can promise you timely review. Some larger involvement will be harder for me since I'm pretty busy at work at this time and will remain so for next couple months.

Also my current gperftools-related work is to finish partially-inlined malloc stuff here: https://github.com/alk/gperftools/commit/2d1ac587c43440596fbcfdef3c550ced5cfc1a40 Feel free to try it btw. If your app spends significant amount of time in constant-sized mallocs/frees (e.g. as happens during stl maps/sets changes or massive protobuf {de,}serializations) it should give you slight boost.

Most of the free list idea seems to be quite little amount of work. But I think we might have to add some more work on top to deal with reclaiming memory out of heaps on free list. That other work might end up  more experimental/researchy in nature and thus end up being large.

acmorrow:
Hi @alk - just circling back to this. Did you make any progress on your free-heap-list idea? Should I pick this work up again and try to move it forward? I'm hoping that we have some solution in place for gperftools-2.6, which I'm guessing is coming up soon-ish?

alk:
Hi Andrew, I actually have _not_ volunteered for implementing this idea. I'm still busy with other stuff. Sorry if I haven't expressed this clearly.

So feel free to grab this idea and implement it or something else that works.

acmorrow:
@alk - Got it. Sorry for the misunderstanding. I also have been busy with other things. Do you think there is enough merit to this PR as is to continue with it (basically, adding tests), or do you think it would be better to abandon it and focus on the heap free list idea?

This is sort of important to us over at MongoDB because we are currently using a modified copy of gperftools embedded in our tree. We would like to get away from that, as it makes people packaging for distributions that encourage use of system packages kind of unhappy. Right now if you build against the system gperftools instead, you lose the optimizations we have made around getting the size of the thread cache.

Also, not pushing, but is there a rough idea when a 2.6 might appear? I'm curious about whether or not it seems possible to get this (or another idea) committed in time for 2.6 or not.

alk:
I'm fine with merging this patch. Regarding release timeline, I have no strict time bounds. I think we've accumulated some amount of good stuff so we shouldn't delays things much. But I'm flexible on waiting a bit more.

acmorrow:
Let me get some quick tests written. I'll push an update tonight or tomorrow

acmorrow:
@alk - I rebased this and added a simple test that calls both functions. I couldn't see a way to test whether the thread still had its cache data structures after calling `MarkThreadTemporarilyIdle` without adding more new functions. LMK if you think that is a good idea, or if you have an idea about how to test that.

Also, we need to do a little internal testing with this patch to double check that we get the same performance effects as our current local modifications, which are slightly different. @Machyne will be doing that testing and can give a yes or no on whether the results are as we expect.

alk:
This is good enough. Thanks a lot. Merged. Let me know if any further fixes need to be applied.

acmorrow:
Thanks!

Issue #409, opened 2015-08-23
Profiling timer always armed on initialization


alk:
Originally reported on Google Code with ID 406
```
The code currently arms the profiling timer unconditionally in ProfileHandler::RegisterThread
for timer sharing detection. Unfortunately, this means the profiling timer will stay
active, even if profiling is inactive and we have shared timers in case if no other
threads get registered later.

This poses a problem for Chromium, which links the profiling code into the executable.
See http://code.google.com/p/chromium/issues/detail?id=115149
```

Reported by `mnissler@google.com` on 2012-02-21 20:50:07

bsilver8192:
Does anybody know what the current status on this is? I would like this issue fixed, and would be happy to take over on getting it merged.

I think @saittam is the original reporter. Any chance you're still interested in working on this, and if not, do you mind if I start with what you did?

saittam:
I'm the original reporter indeed. Unfortunately, this has lingered in the hardly-will-ever-get-to portion of my TODO list since forever. I originally got interested in this because the continuously firing signals tended to confuse gdb (which is a different bug). I've since carried the patch that's attached to the issue in my working trees, but there was nothing else that provided a strong motivation to invest time into pushing for a resolution. At this point, I can't tell whether the problem still exists, and whether the patch still applies... sorry. The original reproduction instructions should still be valid though, so if you're seeing SIGPROF firing in a single-threaded test program after calling ProfileHandler::RegisterThread, the issue is probably still valid.

My lack of spare time to spend is only going to get worse in the upcoming months, so I'd be most delighted if you were to pick this issue up and resolve it.

bsilver8192:
Forgot to include "Fixes: #409" in my PR #762 that just got merged which fixes this... Somebody want to close it manually?

alk:
Fixed by https://github.com/gperftools/gperftools/pull/762. Thanks to Brian Silverman's great effort.

PR #762, opened 2016-01-22
Don't always arm the profiling timer.

bsilver8192:
It causes a noticeable performance hit and can sometimes confuse GDB.

Tested with and without CPUPROFILE_PER_THREAD_TIMERS=1, and on a
linuxthreads system (CPUPROFILE_PER_THREAD_TIMERS=1 isn't supported
there) (I couldn't get all the tests to run there before, but
profile_handler_unittest and profiler_unittest.sh passed).

Originally written by mnissler@google.com. I mostly had to deal with
integrating with CPUPROFILE_PER_THREAD_TIMERS.

Fixes gperftools/gperftools#409

### DIFF
--- a/src/profile-handler.cc
+++ b/src/profile-handler.cc
@@ -79,39 +79,44 @@ struct ProfileHandlerToken {
   void* callback_arg;
 };
 
+// Blocks a signal from being delivered to the current thread while the object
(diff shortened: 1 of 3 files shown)


alk:
Thanks a lot. Looks good. Applied.

PR #765, opened 2016-01-27
fix build with musl libc

### LINKED ISSUE #693: Failed to build with lib musl
Originally reported on Google Code with ID 690

```
What steps will reproduce the problem?
1. unpack source onto x86_64, musl based system
2. apply patch to define __off64_t
3. try to build

What is the expected output? What do you see instead?
Build fails instead of normal completition:
src/malloc_hook_mmap_linux.h: In function 'void* mmap(void*, size_t, int, int, int,
off_t)':
src/malloc_hook_mmap_linux.h:169:18: error: redefinition of 'void* mmap(void*, size_t,
int, int, int, off_t)'
 extern "C" void* mmap(void *start, size_t length, int prot, int flags,
                  ^
In file included from src/malloc_hook.cc:41:0:
src/malloc_hook_mmap_linux.h:155:18: error: 'void* mmap(void*, size_t, int, int, int,
__off64_t)' previously defined here
 extern "C" void* mmap64(void *start, size_t length, int prot, int flags,
                  ^
Makefile:4515: recipe for target 'src/libtcmalloc_minimal_internal_la-malloc_hook.lo'
failed
make: *** [src/libtcmalloc_minimal_internal_la-malloc_hook.lo] Error 1


What version of the product are you using? On what operating system?
Alpine linux 3.0
gcc 4.8.2
musl 1.1.4

Please provide any additional information below.
1. _off64_t is not defined, so I patched base/linux_syscall_support.h to define it
into off64_t
2. musl sys/mman.h is quite different from glibc, so I guess the problem of redifinition
comes from that fact.

I would appreciate for any advice how it could be fixed. Full configure & make logs
attached
```

Reported by `filipp.andronov` on 2015-05-15 17:59:04

<hr>
- _Attachment: [build.log](https://storage.googleapis.com/google-code-attachments/gperftools/issue-690/comment-0/build.log)_
- _Attachment: [10-define-off64-t.patch](https://storage.googleapis.com/google-code-attachments/gperftools/issue-690/comment-0/10-define-off64-t.patch)_

ncopa:
- use off64_t instead of the internal __off64_t
- during configure time, check for __sbrk function. musl libc does not
  have this. sbrk on musl libc will only work when called with 0 as
  argument, so we just let it out for now.
- only do the glibc 32-bit ABI check for mmap/mmap64 on gnu libc. musl
  does not support the 32-bit ABI.

fixes #693

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -163,6 +163,7 @@ AC_HEADER_STDC
 AC_CHECK_TYPES([__int64])       # defined in some windows platforms
 AC_CHECK_TYPES([struct mallinfo],,, [#include <malloc.h>])
 AC_CHECK_TYPES([Elf32_Versym],,, [#include <elf.h>])   # for vdso_support.h
+AC_CHECK_FUNCS(__sbrk)		# for interception on glibc
 AC_CHECK_FUNCS(sbrk)            # for tcmalloc to get memory
 AC_CHECK_FUNCS(geteuid)         # for turning off services when run as root
 AC_CHECK_FUNCS(fork)            # for the pthread_atfork setup
--- a/src/base/linux_syscall_support.h
+++ b/src/base/linux_syscall_support.h
@@ -2166,9 +2166,9 @@ struct kernel_stat {
                          int,                     t, int,       p)
   #endif
   #if defined(__x86_64__)
-    /* Need to make sure __off64_t isn't truncated to 32-bits under x32.  */
+    /* Need to make sure off64_t isn't truncated to 32-bits under x32.  */
     LSS_INLINE void* LSS_NAME(mmap)(void *s, size_t l, int p, int f, int d,
-                                    __off64_t o) {
+                                    off64_t o) {
       LSS_BODY(6, void*, mmap, LSS_SYSCALL_ARG(s), LSS_SYSCALL_ARG(l),
                                LSS_SYSCALL_ARG(p), LSS_SYSCALL_ARG(f),
                                LSS_SYSCALL_ARG(d), (uint64_t)(o));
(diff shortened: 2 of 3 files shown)


alk:
Thanks a lot. I'll take a look this Saturday. Unless someone else manages to review your code sooner.

I'm assuming you've tested in on musl and glibc. I won't test on musl.

ncopa:
I have only compile tested it on musl.

I don't have glibc system handy available, but I expect it to work. It should not break build with uclibc either.

The only thing I am a bit in doubt with is if glibc has `off64_t` or what header needs to be included for it. I thought of adding a configure test for it but thought it is not worse than using `__off64_t`.

plopresti:
I do not think there is any "off64_t" in 64-bit glibc. On 64-bit platforms, it's just called "off_t".

On 32-bit platforms, you have to include <sys/types.h> and, if memory serves, define _FILE_OFFSET_BITS=64 to get the typedef. Or maybe I have that backwards... I have not used 32-bit for some time.

In any event, off64_t is non-standard and arguably not a good idea to use at all...

alk:
IMHO just int64_t is clearest.

Also this makes me wonder whether it makes sense at all to consider patch
that is only compile-tested.

I'm going to test on glibc anyways, but I won't bother with musl for
ethical reasons. I'd prefer to get at least some assurance that this patch
improves thing.

alk:
It turns out docker makes it easy enough to test building it on musl-based alpine linux. This patch compiles but tests fail. By quickly looking at it, I'm seeing that tests fail by expecting strdup to call our malloc while musl's strdup actually calls it's own malloc.

Which is a show stopper at least for dynamically linked case since strdup's result is supposed to be passed to free. Which will be tcmalloc's free and thus fail spectacularly.

I think until it is resolved (or unless there are good reasons to merge regardless e.g. for statically linked case) this patch cannot be merged.

alk:
BTW musl is now officially supporting other malloc implementations interposing over it's own malloc. And found it working when I tested semi-recently. You still need this patch?

PR #766, opened 2016-01-30
Fix include of malloc_hook_c.h in malloc_hook.h

eecsninja:
malloc_hook.h includes malloc_hook_c.h as \<gperftools/malloc_hook_c.h\>.
This requires the compiler to have designated src/ as a standard
include directory (-I), which may not always be the case.

Instead, include it as "malloc_hook_c.h", which will search in the same
directory first. This will always work, regardless of whether src/ was
designated a standard include directory.

### DIFF
--- a/src/gperftools/malloc_hook.h
+++ b/src/gperftools/malloc_hook.h
@@ -70,7 +70,7 @@
 #include <stddef.h>
 #include <sys/types.h>
 extern "C" {
-#include <gperftools/malloc_hook_c.h>  // a C version of the malloc_hook interface
+#include "malloc_hook_c.h"  // a C version of the malloc_hook interface
 }
 
 // Annoying stuff for windows -- makes sure clients can import these functions


alk:
I don't think I agree with your logic about src/ but merged anyways since It won't hurt.

alk:
Thanks for contribution btw.

And I'm curious about your use-case and what was broken for you. Feel free to elaborate.

eecsninja:
See here:
https://codereview.chromium.org/1607303002/diff/80001/base/allocator/allocator_extension.cc

In the Chromium code, there is a copy of gperftools under
src/third_party/tcmalloc/chromium/. Its include files under gperftools/ are
included by other modules in the codebase using the full path shown in the
patch above.

When I attempted to include malloc_hook.h, I got that file not found
message about <gperftools/malloc_hook_c.h>.

alk:
Thanks.

Issue #761, opened 2016-01-22
gperftools is not working for s390x machine


abcshravan:
Why it is not supporting s390x? 
[root@svl gperftools-2.4]# make
/bin/sh ./libtool  --tag=CXX   --mode=compile g++ -DHAVE_CONFIG_H -I. -I./src  -I./src   -DNO_TCMALLOC_SAMPLES -pthread -DNDEBUG -Wall -Wwrite-strings -Woverloaded-virtual -Wno-sign-compare -fno-builtin-malloc -fno-builtin-free -fno-builtin-realloc -fno-builtin-calloc -fno-builtin-cfree -fno-builtin-memalign -fno-builtin-posix_memalign -fno-builtin-valloc -fno-builtin-pvalloc  -Wno-unused-result   -g -O2 -MT src/libtcmalloc_minimal_la-tcmalloc.lo -MD -MP -MF src/.deps/libtcmalloc_minimal_la-tcmalloc.Tpo -c -o src/libtcmalloc_minimal_la-tcmalloc.lo `test -f 'src/tcmalloc.cc' || echo './'`src/tcmalloc.cc
libtool: compile:  g++ -DHAVE_CONFIG_H -I. -I./src -I./src -DNO_TCMALLOC_SAMPLES -pthread -DNDEBUG -Wall -Wwrite-strings -Woverloaded-virtual -Wno-sign-compare -fno-builtin-malloc -fno-builtin-free -fno-builtin-realloc -fno-builtin-calloc -fno-builtin-cfree -fno-builtin-memalign -fno-builtin-posix_memalign -fno-builtin-valloc -fno-builtin-pvalloc -Wno-unused-result -g -O2 -MT src/libtcmalloc_minimal_la-tcmalloc.lo -MD -MP -MF src/.deps/libtcmalloc_minimal_la-tcmalloc.Tpo -c src/tcmalloc.cc  -fPIC -DPIC -o src/.libs/libtcmalloc_minimal_la-tcmalloc.o
In file included from src/tcmalloc.cc:117:0:
src/base/basictypes.h:361:5: error: #error Could not determine cache line length - unknown architecture
 #   error Could not determine cache line length - unknown architecture

zatrazz:
IMHO the CACHELINE_ALIGNED is mainly an optimization hint and should not break build in an unknown configuration. I think instead of emitting an '#error" we should emit a warning and define as if !HAVE___ATTRIBUTE__.

acmorrow:
[...]

PR #767, opened 2016-02-05
Add support for Linux s390x

bryanpkc:
This patch adds support for Linux s390x to gperftools, and resolves issue #761.

libunwind has not yet been ported to the platform, so by default configure enables glibc backtrace. I may add a back chain walker (stacktrace_s390x-inl.h) at a later date, but it would only be useful with a configure option like --enable-s390x-backchain (similar to --enable-frame-pointers), since GCC does not store the back chain on the stack unless the -mbackchain option is used.

[A complete port of linux_syscall_support.h](https://github.com/linux-on-ibm-z/linux-syscall-support) is available.

This patch applies cleanly without @alk's wip-emergency-malloc-explicit branch, so I feel it is time for a pull request.

s390 support is intentionally omitted, as the 31-bit variant is generally being deprecated.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -51,6 +51,13 @@ case "$host" in
    *-darwin*) default_enable_heap_checker=no;;
 esac
 
+# Currently only backtrace works on s390x.
+AC_COMPILE_IFELSE([AC_LANG_PROGRAM(, [return __s390x__])],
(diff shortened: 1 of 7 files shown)


alk:
Thanks a lot! Applied.

PR #772, opened 2016-03-02
Fix various pieces of undefined behavior

bsilver8192:
I caught these running the gperftools tests with [ubsan](http://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html) and maybe [msan](http://clang.llvm.org/docs/MemorySanitizer.html) (don't remember if ubsan or msan caught uninitialized values first).

### DIFF
--- a/src/base/low_level_alloc.cc
+++ b/src/base/low_level_alloc.cc
@@ -107,7 +107,7 @@ static int IntLog2(size_t size, size_t base) {
 
 // Return a random integer n:  p(n)=1/(2**n) if 1 <= n; p(n)=0 if n < 1.
 static int Random() {
-  static int32 r = 1;         // no locking---it's not critical
+  static uint32 r = 1;         // no locking---it's not critical
   ANNOTATE_BENIGN_RACE(&r, "benign race, not critical.");
   int result = 1;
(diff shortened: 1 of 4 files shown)


alk:
Thanks. Not sure about last commit, yet, since it introduces dependency on C++11 feature I think.

bsilver8192:
Are you talking about the variadic macro? I didn't think about it before, but looking around it appears that isn't even standard C++11; it's a GCC extension... [This Stack Overflow question](http://stackoverflow.com/questions/5588855/standard-alternative-to-gccs-va-args-trick) basically says there isn't a standard way of doing it. How about just declaring the `IncrementCallsTo##hook_type` functions separately?

alk:
Yes. Feel free to fix it. I'll take all commits before last.

alk:
Actually C99 variadic macros are support by msvc 2013 and older gccs.

So that's workable approach I think.

alk:
Leaving it up to you anyways.

bsilver8192:
Sounds good to me. I was ignoring that because C99 variadic macros have no good way to handle "no arguments", but then I realized none of those functions actually have that problem... Done

alk:
Thanks a lot. All applied.

PR #776, opened 2016-03-08
Recognize modern Linux ARM

bsilver8192:
Commit e580d7888154fa7f95b3cef9e18f1ce69182212b fixed the macros in some
of the code but not other places.

`make check` still fails in the same places on a Debian Jessie armhf
system.


alk:
Applied. Thanks a lot.

bsilver8192:
Thank you too!

PR #785, opened 2016-03-22
Size class statistics API

ksuarz:
**TL;DR**: I think exposing a size class API is valuable despite the discussion in #746.

This patch adds an API for extracting information about central freelist size classes beyond
what is already provided in `DumpStats()`. A callback can be registered to extract statistics from `SizeClasses` on a per-size class basis, which is somewhat similar to how `Ranges` works.

In #746, there are some questions as to why such an API would be useful. We're [investigating TCMalloc fragmentation issues in MongoDB](https://groups.google.com/d/topic/gperftools/gilTxIMuzqY/discussion) and it is generally useful to see which size classes are experiencing fragmentation and where the memory is being held (central, transfer, thread cache).

Some observations:
- `DumpStats()` provides some size class information but does not go into this level of detail. It also exposes this is a giant text blob, which is inefficient to both generate and parse.
- I acknowledge that a struct is not the most portable way to expose the statistics, but it is efficient. I could add a version number to it at the cost of some elegance.
- I understand that this is a TCMalloc-specific detail and not a generic malloc statistic. The default implementation does nothing, to support implementations where there is no notion of size classes.

### DIFF
--- a/src/central_freelist.h
+++ b/src/central_freelist.h
@@ -74,6 +74,12 @@ class CentralFreeList {
   // Returns the number of free objects in the transfer cache.
   int tc_length();
 
+  // Returns the number of spans in both the empty and nonempty freelists.
+  int spans() {
+    SpinLockHolder h(&lock_);
+    return num_spans_;
+  }
+
   // Returns the memory overhead (internal fragmentation) attributable
   // to the freelist.  This is memory lost when the size of elements
   // in a freelist doesn't exactly divide the page-size (an 8192-byte
--- a/src/gperftools/malloc_extension.h
+++ b/src/gperftools/malloc_extension.h
@@ -69,6 +69,7 @@ typedef std::string MallocExtensionWriter;
 
 namespace base {
 struct MallocRange;
+struct MallocSizeClass;
 }
 
 // Interface to a pluggable system allocator.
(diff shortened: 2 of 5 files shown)


alk:
Thanks a lot. This looks good. I'll need tiny test to that exercises new code at least a tiny bit so that I can be sure that it doesn't crash at least.

ksuarz:
Hey, sorry for the radio silence on this. I've been quite busy but I can whip up some tests sometime soon after our big development cycle winds down. I'm assuming that malloc_extension_test is the right place to put it?

So far, this code has been running well in the bundled TCMalloc in a development version of MongoDB, so I'm somewhat confident it's in decent shape :)

acmorrow:
I think it would be valuable to try to get this merged before the next release of gperftools. @ksuarz, could you add the test that @alk suggested?

ksuarz:
Hi @alk, I've added a tiny unit test akin to the Ranges unit test and rebased both commits onto master. Please let me know what you think.

acmorrow:
Ping @alk - Per some other discussion, it sounds like a new version is nearing, and it would be good to see this merged before that if you think it is ready.

ksuarz:
Hey @alk, not to be a bother, but could you please take a look at this when you get the chance? I'm happy to improve the test in whatever way possible to get this in.

alk:
Apologies for delay. Can you please:

a) drop extra blank line introduced by first commit in malloc_extension.cc before SetNumericProperty

b) add comments for fields of struct MallocSizeClass 

c) change unittest to be a bit more meaningful? Perhaps I am missing something but malloc-ing 1 meg in your test doesn't seem to matter at all. Perhaps something like, on first pass check invariants and note stats for some particular size class. Then allocate something from that size class and observe how things changed?

d) add something matching to malloc_extension_c.h

alk:
Closing due to inactivity. But feel free to reopen if/when you're ready to move this forward.

Issue #775, opened 2016-03-08
Unsafe call made during signal handling - ProfileHandler::SignalHandler and GetStackTraceWithContext


jeking3:
I've been looking at a core on an Ubuntu 14.04.3 LTS system where SIGABRT was delivered to the application.  The top level libprofiler signal handler delivers a PTRACE_KILL to each thread.  This invoked the following stack which deadlocked for hopefully obvious reasons:

Thread 226 (Thread 0x7fc26fda7700 (LWP 3331)):
#0  __lll_lock_wait () at ../nptl/sysdeps/unix/sysv/linux/x86_64/lowlevellock.S:135
#1  0x00007fcdc8030672 in _L_lock_953 () from ./lib/x86_64-linux-gnu/libpthread.so.0
#2  0x00007fcdc80304da in __GI___pthread_mutex_lock (mutex=0x7fcdcb144990 <_rtld_global+2352>) at ../nptl/pthread_mutex_lock.c:114
#3  0x00007fcdca19db5f in __GI___dl_iterate_phdr (callback=0x7fcdbcdea500, data=0x7fc26fd9c430) at dl-iteratephdr.c:41
#4  0x00007fcdbcdeaade in ?? () from ./usr/lib/x86_64-linux-gnu/libunwind.so.8
#5  0x00007fcdbcde8bb5 in ?? () from ./usr/lib/x86_64-linux-gnu/libunwind.so.8
#6  0x00007fcdbcde9db9 in ?? () from ./usr/lib/x86_64-linux-gnu/libunwind.so.8
#7  0x00007fcdbcdea4e9 in ?? () from ./usr/lib/x86_64-linux-gnu/libunwind.so.8
#8  0x00007fcdbcde6a31 in _ULx86_64_step () from ./usr/lib/x86_64-linux-gnu/libunwind.so.8
#9  0x00007fcdc2b98d54 in GetStackTraceWithContext(void*_, int, int, void const_) () from ./usr/lib/libprofiler.so.0
#10 0x00007fcdc2b968ce in ?? () from ./usr/lib/libprofiler.so.0
#11 0x00007fcdc2b973f3 in ProfileHandler::SignalHandler(int, siginfo_t_, void_) () from ./usr/lib/libprofiler.so.0
#12 <signal handler called>
[...]

PR #790, opened 2016-04-06
A bug fix attempt for issue #775

jingleyang:
Hi  @alk ,

Could you please give me some feedback about this bug fix attempt for issue #775 . 

The more detailed report can be found at here  [https://github.com/jingleyang/security_ctf/blob/master/opensource/A%20deadlock%20bug%20fix%20solution%20for%20gperftools%20issues%20775.pdf](https://github.com/jingleyang/security_ctf/blob/master/opensource/A%20deadlock%20bug%20fix%20solution%20for%20gperftools%20issues%20775.pdf)

Thank you for your time.

Cheers,
Jingle

### DIFF
--- a/src/stacktrace.cc
+++ b/src/stacktrace.cc
@@ -243,13 +243,30 @@ namespace tcmalloc {
 namespace {
   using tcmalloc::EnterStacktraceScope;
   using tcmalloc::LeaveStacktraceScope;
+  // A patch for isuess 775. https://github.com/gperftools/gperftools/issues/775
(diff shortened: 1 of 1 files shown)


alk:
Hi. Thanks for the patch.

I think we cannot do __thread variables in the code that is shared between different platforms. We could try doing it for libgcc unwinder.

But note it won't 100% work anyways. Which is known issue with this unwinder (perhaps I should not make it available by default).

I.e. consider what happens when the program is unwinding stack due to C++ exception (using same libgcc runtime) and we get profiling signal and try to capture backtrace too. Will get same lockup even with this kind of protection.

So because this is not fixing fundamental issue with libgcc unwinder I think I will not take your patch.

Again, huge thanks for looking into this.

PR #793, opened 2016-04-15
Small performance tweak: avoid calling time() if we don't need it

vertexodessa:
time() is a system call, which might lead to performance overhead.

### DIFF
--- a/src/heap-profiler.cc
+++ b/src/heap-profiler.cc
@@ -272,7 +272,7 @@ static void MaybeDumpProfileLocked() {
     const int64 inuse_bytes = total.alloc_size - total.free_size;
     bool need_to_dump = false;
     char buf[128];
-    int64 current_time = time(NULL);
+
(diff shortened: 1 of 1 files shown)


src/heap-profiler.cc
```diff
@@ -293,13 +293,15 @@ static void MaybeDumpProfileLocked() {
...
       need_to_dump = true;
-    } else if (FLAGS_heap_profile_time_interval > 0 &&
-               current_time - last_dump_time >=
-               FLAGS_heap_profile_time_interval) {
-      snprintf(buf, sizeof(buf), "%" PRId64 " sec since the last dump",
-               current_time - last_dump_time);
-      need_to_dump = true;
-      last_dump_time = current_time;
+    } else if (FLAGS_heap_profile_time_interval > 0 ) {
+      int64 current_time = time(NULL);
```
jingleyang: +1

alk:
Applied. Thanks.

PR #806, opened 2016-06-10
Hugetlb mmap allocator

Lastique:
Add a new hugetlb_mmap system allocator. The new allocator allows to back the allocated memory with huge pages. Unlike the memfs allocator, hugetlb_mmap does not require a mounted hugetlbfs filesystem and also allows to select the huge page size on systems that support multiple huge page sizes. It also does not require an open file descriptor for memory allocation.

Not requiring hugetlbfs mount point reduces administration burden on systems where hugetlbfs is not mounted by default (e.g. on Debian 7). Also, as I understand, using mmap directly allows to use the dynamic allocation of huge pages, when possible, while hugetlbfs only uses the preallocated pool.

The new allocator is Linux-specific, I'm not sure other systems have support for `MAP_HUGETLB` and `MAP_HUGE_SHIFT`.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -472,6 +472,7 @@ libtcmalloc_minimal_internal_la_SOURCES = src/common.cc \
                                           src/internal_logging.cc \
                                           $(SYSTEM_ALLOC_CC) \
                                           src/memfs_malloc.cc \
+                                          src/hugetlb_mmap_malloc.cc \
                                           src/central_freelist.cc \
                                           src/page_heap.cc \
                                           src/sampler.cc \
--- a/docs/tcmalloc.html
+++ b/docs/tcmalloc.html
@@ -390,7 +390,7 @@ threads. The raw data used to generate these graphs (the output of the
 </table>
 
 
-<ul> 
+<ul>
   <li> TCMalloc is much more consistently scalable than PTMalloc2 - for
        all thread counts &gt;1 it achieves ~7-9 million ops/sec for small
        allocations, falling to ~2 million ops/sec for larger
@@ -609,6 +609,61 @@ tries to allocate memory from the kernel.</p>
   <td>default: false</td>
   <td>
      If true, use MAP_PRIVATE when mapping via memfs, not MAP_SHARED.
+  </td>
+</tr>
+
+<tr valign=top>
+  <td><code>TCMALLOC_HUGETLB_MMAP_ENABLE</code></td>
+  <td>default: false</td>
+  <td>
+     If true, try allocating huge pages from the system. Note that
+     unlike the memfs allocator, this allocator does not require
+     hugetlbfs filesystem to be mounted.
+  </td>
+</tr>
+
+<tr valign=top>
+  <td><code>TCMALLOC_HUGETLB_MMAP_PAGE_SIZE_KB</code></td>
+  <td>default: 0</td>
+  <td>
+     The desired huge page size to use when requesting memory from system,
+     in KiB. The allocator will select the huge page size out of the page
+     sizes supported by the system so that the selected size does not exceed
+     this value. If 0 or unset, the default huge page size will be used.
+  </td>
+</tr>
+
+<tr valign=top>
+  <td><code>TCMALLOC_HUGETLB_MMAP_LIMIT_MB</code></td>
+  <td>default: 0</td>
+  <td>
+     Limit total hugetlb_mmap allocation size to specified number of MiB.
+     0 means "no limit".
+  </td>
+</tr>
+
+<tr valign=top>
+  <td><code>TCMALLOC_HUGETLB_MMAP_LIMIT_MB</code></td>
+  <td>default: 0</td>
+  <td>
+     Limit total hugetlb_mmap allocation size to specified number of MiB.
+     0 means "no limit".
+  </td>
+</tr>
+
+<tr valign=top>
+  <td><code>TCMALLOC_HUGETLB_MMAP_ABORT_ON_FAIL</code></td>
+  <td>default: false</td>
+  <td>
+     If true, abort() whenever hugetlb_mmap allocator fails to satisfy an allocation.
+  </td>
+</tr>
+
+<tr valign=top>
+  <td><code>TCMALLOC_HUGETLB_MMAP_IGNORE_MMAP_FAIL</code></td>
+  <td>default: false</td>
+  <td>
+     If true, ignore failures from mmap in the hugetlb_mmap allocator.
   </td>
 </tr>
 
--- a/src/hugetlb_mmap_malloc.cc
+++ b/src/hugetlb_mmap_malloc.cc
@@ -1,0 +1,428 @@
+// -*- Mode: C++; c-basic-offset: 2; indent-tabs-mode: nil -*-
+// Copyright (c) 2016, gperftools Contributors
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are
+// met:
+//
+//     * Redistributions of source code must retain the above copyright
+// notice, this list of conditions and the following disclaimer.
+//     * Redistributions in binary form must reproduce the above
+// copyright notice, this list of conditions and the following disclaimer
+// in the documentation and/or other materials provided with the
+// distribution.
+//     * Neither the name of Google Inc. nor the names of its
+// contributors may be used to endorse or promote products derived from
+// this software without specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+// "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+// LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+// A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+// OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+// SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+// LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+// DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+// THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+// (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+// ---
+// Author: Andrey Semashev
+//
+// A tcmalloc system allocator that uses mmap with MAP_HUGETLB flag.
+//
+// Since it only exists on linux, we only register this allocator there.
+
+#ifdef __linux
+
+#include <config.h>
+#include <errno.h>                      // for errno
+#include <stddef.h>                     // for size_t, NULL
+#ifdef HAVE_STDINT_H
+#include <stdint.h>                     // for uintptr_t
+#endif
+#include <string.h>                     // for strerror, memmove, memchr, strncmp
+#include <sys/mman.h>                   // for mmap, MAP_FAILED, etc
+
+#if defined(MAP_HUGETLB)
+
+#include <sys/types.h>
+#include <sys/stat.h>                   // for stat, fstat
+#include <fcntl.h>                      // for open
+#include <unistd.h>                     // for close, read
+#include <new>                          // for operator new
+#include <algorithm>                    // for lower_bound
+
+#include <gperftools/malloc_extension.h>
+#include "base/basictypes.h"
+#include "base/googleinit.h"
+#include "internal_logging.h"
+
+DEFINE_bool(hugetlb_mmap_malloc_enable,
+            EnvToBool("TCMALLOC_HUGETLB_MMAP_ENABLE", false),
+            "Enable hugetlb_mmap system memory allocator.");
+#if defined(MAP_HUGE_SHIFT)
+DEFINE_uint64(hugetlb_mmap_malloc_page_size_kb,
+              EnvToInt64("TCMALLOC_HUGETLB_MMAP_PAGE_SIZE_KB", 0),
+              "Specifies the preferred page size in KiB. "
+              "The allocator will choose the closest available page size "
+              "not greater than the specified value.  0 == use the default "
+              "huge page size.");
+#endif
+DEFINE_uint64(hugetlb_mmap_malloc_limit_mb,
+              EnvToInt64("TCMALLOC_HUGETLB_MMAP_LIMIT_MB", 0),
+              "Limit total allocation size to the "
+              "specified number of MiB.  0 == no limit.");
+DEFINE_bool(hugetlb_mmap_malloc_abort_on_fail,
+            EnvToBool("TCMALLOC_HUGETLB_MMAP_ABORT_ON_FAIL", false),
+            "abort() whenever hugetlb_mmap_malloc fails to satisfy an allocation "
+            "for any reason.");
+DEFINE_bool(hugetlb_mmap_malloc_ignore_mmap_fail,
+            EnvToBool("TCMALLOC_HUGETLB_MMAP_IGNORE_MMAP_FAIL", false),
+            "Ignore failures from mmap");
+
+namespace tcmalloc {
+namespace {
+
+// Hugetlb+mmap based allocator for tcmalloc
+class HugetlbMmapSysAllocator: public SysAllocator {
+private:
+  struct HugepageInfo {
+    const char* sysfs_dir_name;
+    size_t page_size;
+  };
+
+public:
+  explicit HugetlbMmapSysAllocator(SysAllocator* fallback, size_t page_size)
+    : fallback_(fallback),
+      page_size_(page_size),
+      allocated_size_(0u),
+      mmap_flags_(MAP_PRIVATE | MAP_ANONYMOUS | MAP_HUGETLB
+#if defined(MAP_HUGE_SHIFT)
+        | (get_log2(page_size) << MAP_HUGE_SHIFT)
+#endif
+      ),
+      alloc_failed_(false) {
+  }
+
+  void* Alloc(size_t size, size_t* actual_size, size_t alignment);
+  static void Initialize();
+
+private:
+  static bool ParseMeminfo(int meminfo_fd, size_t* page_size);
+
+#if defined(MAP_HUGE_SHIFT)
+  static unsigned int get_log2(uint64 n) {
+    if (n & 0x00000000FFFFFFFFull)
+      return get_log2(static_cast< uint32 >(n));
+    else
+      return get_log2(static_cast< uint32 >(n >> 32)) + 32u;
+  }
+
+  static unsigned int get_log2(uint32 n) {
+    if (n & 0x0000FFFFu)
+      return get_log2(static_cast< uint16 >(n));
+    else
+      return get_log2(static_cast< uint16 >(n >> 16)) + 16u;
+  }
+
+  static unsigned int get_log2(uint16 n) {
+    if (n & 0x00FFu)
+      return get_log2(static_cast< uint8 >(n));
+    else
+      return get_log2(static_cast< uint8 >(n >> 8)) + 8u;
+  }
+
+  static unsigned int get_log2(uint8 n) {
+    unsigned int res = 0u;
+    if ((n & 0x0Fu) == 0u) {
+      res += 4u;
+      n >>= 4;
+    }
+    if ((n & 0x03u) == 0u) {
+      res += 2u;
+      n >>= 2;
+    }
+    if ((n & 0x01u) == 0u) {
+      res += 1u;
+    }
+
+    return res;
+  }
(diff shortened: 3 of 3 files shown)


alk:
Hi Andrey. Thanks for the patch. And sorry for delay.

Few comments:

* I don't think we need complications around log2. I.e. gnuc and non-gnuc versions and all those overloads. I.e. I'd just do one generic portable uint64_t version (and I'd convert arg to all 1s after highest bit followed by portable popcount; I think this is how hacker's delight proposes doing it). Or I think we can even require page size arg to be power of two in which case it is just popcount(arg-1).

* is it possible to unify this with memfs allocator or just generic system-alloc? I think that would be great thing to have instead of introducing somewhat duplicated logic.

* do we really need to do all this logic of stat-ing sysfs and parsing meminfo? My thinking is given that this requires users to reserve the memory anyways, should should know what are available page sizes and how to configure everything. Thus I don't think we need any of logic to detect availability of of various sizes and picking right size class. I'd just require user to set right environment vars.

What you think ?

Lastique:
> I don't think we need complications around log2. I.e. gnuc and non-gnuc versions and all those overloads. I.e. I'd just do one generic portable uint64_t version (and I'd convert arg to all 1s after highest bit followed by portable popcount; I think this is how hacker's delight proposes doing it).

The only portable function related to this is [ffs](http://pubs.opengroup.org/onlinepubs/009695399/functions/ffs.html), and it's only for `int`s. glibc also offers [overloads](http://man7.org/linux/man-pages/man3/ffs.3.html) for `long` and `long long` but I don't want to limit the implementation to glibc only, so I would have to implement these myself. We could rely on the fact that the current page sizes fit into `int` andyway, but I don't like leaving such security timebomb in the code.

I know there are [other](https://graphics.stanford.edu/~seander/bithacks.html#CountBitsSet64) implementations of the similar functionality, without compiler builtins, but they all seem more complicated and slower (`__builtin_ctz` is a single fast instruction compared to the sequences presented there). I don't think a portable popcount (in the absence of the corresponding instruction) would be more efficient and compact than the portable version of `get_log2`, so I don't see much point for reimplementing it through popcount.

However, I acknowledge that this code is not related to the allocator itself and could be moved to generic utilities. If there is a more appropriate place in the code base, I can move it there.

> Or I think we can even require page size arg to be power of two in which case it is just popcount(arg-1).

That argument eventually comes from user settings. I'll expand on this below.

> is it possible to unify this with memfs allocator or just generic system-alloc? I think that would be great thing to have instead of introducing somewhat duplicated logic.

I looked at memfs (`HugetlbSysAllocator` in particular) before implementing this and decided not to modify it. That allocator relies on the file descriptor being used and it should be completely avoided in the new allocator. Arguably, the new allocator could *replace* `HugetlbSysAllocator` if the filesystem-related functionality is not considered useful, but I'm not proposing this as part of this PR.

I took a brief look at system-alloc.cc, and `MmapSysAllocator` in particular. It could be used as a base for the new implementation, but it seems to me that this allocator is not even used by default since `SbrkSysAllocator` comes first in the list and presumably it succeeds as long as there is memory. Or I am missing something.

> do we really need to do all this logic of stat-ing sysfs and parsing meminfo? My thinking is given that this requires users to reserve the memory anyways, should should know what are available page sizes and how to configure everything. Thus I don't think we need any of logic to detect availability of of various sizes and picking right size class. I'd just require user to set right environment vars.

I think we do. First all, I don't trust user's settings. Mismatches can be introduced for different reasons - innocent mistake, malicious intents or even the specifics of the application design and deployment. For example, in my application where I needed this, we package the application with a pre-defined setting of the preferred huge page size, that is reasonably good for the application purposes. But we have no full control over the target systems where the application will be installed, and we don't know exactly how huge pages will be configured there, if at all. We want our application to work top-notch in any case and use huge pages opportunistically.

It is true that the huge pages have to be configured anyway. But they can be configured in different ways, including the default page size and the number of reserved and surplus pages of different sizes.

Lastique:
So, is there anything I should do about this PR?

alk:
I am still unhappy about parsing /proc/meminfo and stating /sys.

How about something simpler such as:

a) if page size is given, just pass it to mmap in flags. If it fails do fallback. No need to hardcode set of supported sizes.

b) if page size is not given, just do MAP_HUGETLB and discover page size at run time. (See below)

On minor things, log2 thing should also be replaced with single portable implementation. It does not have to be fast. Also there is no need to have 3 separate commits since splitting it in 3 doesn't make review or reading history easier.

For probing page size I found that this works (we have to do MAP_FIXED trick because kernel does rounding of length argument up to page size):

```
#include <sys/types.h>
#include <sys/mman.h>
#include <unistd.h>
#include <stdlib.h>
#include <stdio.h>
#include <inttypes.h>

int main()
{
        void *addr1 = mmap(0, 2U << 30, PROT_NONE, MAP_PRIVATE|MAP_ANONYMOUS, 0, 0);
        if (addr1 == MAP_FAILED) {
                perror("mmap(PROT_NONE)");
                abort();
        }
        uintptr_t aligner = (1U << 30) - 1;
        char *addr2 = (char *)(((uintptr_t)addr1 + aligner) & ~aligner);

        uintptr_t shift = 12;
        while (shift < 30) {
                char *addr3 = mmap(addr2 + (1U << shift), 4096,
                                   PROT_READ|PROT_WRITE,
                                   MAP_PRIVATE|MAP_HUGETLB|MAP_FIXED|MAP_ANONYMOUS, 0, 0);
                if (addr3 != MAP_FAILED) {
                        break;
                }
                shift++;
        }
        printf("shift = %llu, page_size = %llu\n", (unsigned long long)shift,
               (unsigned long long)1 << shift);
        return 0;
}
```

alk:
Here is better version of page size probing that actually first ensures that MAP_HUGETLB works. So we can do that as part of first allocation:

```
#include <sys/types.h>
#include <sys/mman.h>
#include <unistd.h>
#include <stdlib.h>
#include <stdio.h>
#include <inttypes.h>

int main()
{
        void *addr1 = mmap(0, 2U << 30, PROT_NONE, MAP_PRIVATE|MAP_ANONYMOUS, 0, 0);
        if (addr1 == MAP_FAILED) {
                perror("mmap(PROT_NONE)");
                abort();
        }
        uintptr_t aligner = (1U << 30) - 1;
        char *addr2 = (void *)(((uintptr_t)addr1 + aligner) & ~aligner);
        char *addr3 = mmap(addr2, 4096,
                           PROT_READ|PROT_WRITE,
                           MAP_PRIVATE|MAP_HUGETLB|MAP_FIXED|MAP_ANONYMOUS, 0, 0);
        if (addr3 == MAP_FAILED) {
                perror("FIXED");
                abort();
        }
        *addr3 = 0;

        /*
         * uintptr_t shift = 12;
         * while (shift < 30) {
         *      size_t s = 1U << shift;
         *      unsigned char p = 0;
         *      int rv = mincore(addr3 + s, 1, &p);
         *      if (rv != 0) {
         *              perror("mincore");
         *              abort();
         *      }
         *      if (p == 0) {
         *              break;
         *      }
         *         shift++;
         * }
         */
        /*
         * uintptr_t shift = 12;
         * while (shift < 30) {
         *      size_t s = 1U << shift;
         *      char *addr4 = mmap(addr3 + s, 4096, PROT_NONE, MAP_PRIVATE|MAP_FIXED|MAP_ANONYMOUS, 0, 0);
         *      if (addr4 != MAP_FAILED) {
         *              break;
         *      }
         *         shift++;
         * }
         */
        uintptr_t shift = 12;
        while (shift < 30) {
                size_t s = 1U << shift;
                int rv = munmap(addr3 + s, addr2 + (1U << 30) - addr3 - s);
                if (rv == 0) {
                        break;
                }
                shift++;
        }
        printf("shift = %llu, page_size = %llu\n", (unsigned long long)shift,
               (unsigned long long)1 << shift);
        return 0;
}
```

Lastique:
@alk I'm not sure how your code is supposed to work - all `addr1`, `addr2` and `addr3` may end up being the same and `munmap` will succeed regardless of the page length. Besides, different architectures have different sizes of huge pages, not just 1 MiB and up, some even multiple sizes. So, unless there is a reliable way to obtain the size of a huge page, I still prefer my current approach.

I don't mind removing the optimized version of `get_log2`, if you want me to.

Please, feel free to close this PR if you find it unacceptable.

alk:
Apologies for confusion here is how this code works. First it allocates 2 gigs of virtual address space. then within this chunk of address space it finds 1Gb-aligned address and allocates single hugetlb page there. It relies on kernel's behavior of rounding size argument up to huge page size. This is at addr2 (and addr3 is same; this is where I've introduced unnecessary confusion).

Then it probes actually allocated length by trying to unmap area just after tried page size. If we're trying too small page size, we'll ask kernel to split huge page and kernel refuses. So it will first succeed on exact page size that is behind MAP_HUGETLB.

While those are two assumptions that are not super-portable, I think they're reasonably sane (more so second). So I believe this is simpler and thus better way to support MAP_HUGETLB then your proposal. In proposal my biggest concern is parsing all the things which has tiny bit of security risk (i.e. if someone runs stuff without /proc and/or /sys mounted) and makes assumption about format of files.

But do let me know what you think.

Lastique:
> Apologies for confusion here is how this code works. First it allocates 2 gigs of virtual address space. then within this chunk of address space it finds 1Gb-aligned address and allocates single hugetlb page there. It relies on kernel's behavior of rounding size argument up to huge page size. This is at addr2 (and addr3 is same; this is where I've introduced unnecessary confusion).

Since you have no control of address selection, the first `mmap` of the 2 GiB can return address that is already 1 GiB aligned, and then the second `mmap` will return the same address again.

Also, I don't think allocating 2 GiB of address space is a good idea in terms of resource usage. Address space is a scarce resource on 32-bit systems and on systems with limited overcommit. Even on 64-bit systems I'm not sure the kernel will be able to allocate that space without reserving the storage for it (in the main memory and/or swap) on every architecture. Note also that multiple processes may be starting and performing the same kind of allocation, which will increase chances of errors.

> In proposal my biggest concern is parsing all the things which has tiny bit of security risk (i.e. if someone runs stuff without /proc and/or /sys mounted) and makes assumption about format of files.

That is a fair concern, and I'd be willing to improve security of the code. For instance, I could add a call to `fstat` before parsing /proc/meminfo to verify that the file is owned by root (uid == 0). In order to forge that the attacker would have to have a root access to the machine or replace `fstat` (e.g. with LD_PRELOAD). In both cases, I believe, there's nothing we can do (i.e. even the current code can't be trusted to operate correctly) and the system is compromised beyond return.

I was thinking about also checking that /proc/meminfo is part of procfs filesystem, but I can't find a way to do that except to parse /etc/mtab, and at that point it is no longer more protected than just parsing /proc/meminfo straight away, let alone the added complexity in the code.

Would the `fstat` check work for you? If you have other suggestions about improving security, please let me know.

Realistically though, since the proposed piece of code is Linux-specific, I think it is safe to assume procfs and sysfs are present. I don't think any distro disables them.

Lastique:
By the way, it should be noted that libhugetlbfs, the official package with utilities for using huge pages, does parse /proc/meminfo for the default huge page size. It also uses /sys/kernel/mm/hugepages to work with multiple huge page sizes. I think, this is *the* interface to work with huge pages.

alk:
>> Since you have no control of address selection, the first mmap of the 2 GiB can return address that is already 1 GiB aligned, and then the second mmap will return the same address again.

And it'll still work fine.

>> Also, I don't think allocating 2 GiB of address space is a good idea in terms of resource usage. Address space is a scarce resource on 32-bit systems and on systems with limited overcommit. Even on 64-bit systems I'm not sure the kernel will be able to allocate that space without reserving the storage for it (in the main memory and/or swap) on every architecture. Note also that multiple processes may be starting and performing the same kind of allocation, which will increase chances of errors.

The code is explicitly doing PROT_NONE for that 2GB mapping. So nothing gets "booked" on committed memory. We may hit address space rlimit in this case, but failing in such case is fine and we'll simply fall back to not using huge pages. As for 32-bit systems we can either probe for smaller pages (say up to 16 megs), or we can just disable it for them completely.

So I still think my proposal would be work great. But libhugetlbfs sounds like great alternative. We should not link to it directly though, to avoid forcing this dependency on everyone. We should use their symbols weakly, so that if libhugetlbfs is linked or LD_PRELOAD-ed then we use it, otherwise we dont.

Lastique:
> And it'll still work fine.

I'm not sure how, since there would be two mappings for the same address.

> But libhugetlbfs sounds like great alternative. We should not link to it directly though, to avoid forcing this dependency on everyone. We should use their symbols weakly, so that if libhugetlbfs is linked or LD_PRELOAD-ed then we use it, otherwise we dont.

I wasn't suggesting to use libhugetlbfs. In fact, I don't think it is possible because it performs dynamic memory allocations while doing its stuff. I mentioned the library because it parses the procfs and sysfs as well and apparently this is the only interface the kernel provides to userland to discover huge page properties.

Lastique:
> The code is explicitly doing PROT_NONE for that 2GB mapping. So nothing gets "booked" on committed memory. We may hit address space rlimit in this case, but failing in such case is fine and we'll simply fall back to not using huge pages. As for 32-bit systems we can either probe for smaller pages (say up to 16 megs), or we can just disable it for them completely.

Hmm, I googled around and didn't find any evidence that `PROT_NONE` influences memory allocation. It probably could, but I suspect the possibility of a subsequent `mprotect` call with a less restricting access mode pretty much prohibits this kind of optimization.

32-bit applications and systems with limited overcommit are still an issue. I don't think allocating 2 GiB of space on start is acceptable for our application, and I'm sure such behavior would be very surprising for other users as well.

alk:
>> Since you have no control of address selection, the first mmap of the 2 GiB can return address that is already 1 GiB aligned, and then the second mmap will return the same address again.

And it'll still work fine.

>> Also, I don't think allocating 2 GiB of address space is a good idea in terms of resource usage. Address space is a scarce resource on 32-bit systems and on systems with limited overcommit. Even on 64-bit systems I'm not sure the kernel will be able to allocate that space without reserving the storage for it (in the main memory and/or swap) on every architecture. Note also that multiple processes may be starting and performing the same kind of allocation, which will increase chances of errors.

The code is explicitly doing PROT_NONE for that 2GB mapping. So nothing gets "booked" on committed memory. We may hit address space rlimit in this case, but failing in such case is fine and we'll simply fall back to not using huge pages. As for 32-bit systems we can either probe for smaller pages (say up to 16 megs), or we can just disable it for them completely.

So I still think my proposal would be work great. But libhugetlbfs sounds like great alternative. We should not link to it directly though, to avoid forcing this dependency on everyone. We should use their symbols weakly, so that if libhugetlbfs is linked or LD_PRELOAD-ed then we use it, otherwise we dont.

alk:
PROT_NONE really only reserves memory and it is portable even beyond Linux. Subsequent mprotect to other protection flags would perform committting of memory under mprotect and may fail.

alk:
Correction, PROT_NONE only reserves address space. And never applies to any committed memory limits. It only affects RLIMIT_AS as noted above.

Lastique:
I've updated the PR:

- Merged all changes into one commit, as requested.
- Removed the gcc-specific version of `get_log2`, only left the portable version.
- Added `stat`/`fstat` checks to ensure the sysfs/procfs entries are not symlinks and are owned by root. Also added `O_NOFOLLOW` to the `open` call.
- Corrected a few comments.

I did not remove the use of sysfs/procfs because I don't think the suggested alternative (the auto-detection through `munmap`) is acceptable for us and I don't see another way to do this.

alk:
Hi Andrei. I still don't like the idea of parsing some sysfs files.

Can you please elaborate on why my alternative is not acceptable for you? As yet another alternative to that, can we do weak dependency on libhugetlbfs instead ? (But I think there is nothing wrong with my first alternative proposal and kinda like it)

Lastique:
> Can you please elaborate on why my alternative is not acceptable for you?

I already mentioned some of my concerns:

- The auto-detection code allocates a huge page, whose number is limited. Another process(es) might be using most or all huge pages and the auto-detection algorithm will fail. This will result in huge pages not being used, even if the other process releases its pages or terminates. This would be a very possible situation with our application, because there are multiple (dozens) processes running, all using huge pages. Also, often these processes start in packs, so it is possible that multiple processes run the detection code concurrently.
- I did not test the auto-detection code on 32-bit systems, systems with 2 GiB or less memory, or systems with limited/disabled overcommit and I'm not sure how the code will perform there. Even though our application is currently 64-bit only, I don't want to restrict this allocator to 64-bit only. Limited overcommit and low memory are supported (in the case of overcommit - even the intended) use cases for our application, so the allocator must perform well in these conditions.
- Lastly, I simply don't like the idea. Allocating 2 GiB of space does not feel like an adequate resource usage to me.

I'd like to stress that I myself am not particularly happy about having to parse procfs, but I just don't see another way to do this. It seems this is the intended interface to obtain this information.

> As yet another alternative to that, can we do weak dependency on libhugetlbfs instead ?

No, I don't think we can.

1. libhugetlbfs requires hugetlbfs mount point. In this regard it offers nothing over what `HugetlbSysAllocator` (memfs_malloc.cc) already provides.
2. libhugetlbfs routines do dynamic memory allocation and use locale while performing. I don't think this is allowed during libtcmalloc initialization.
3. libhugetlbfs obtains page sizes by parsing sysfs and procfs. It will also parse files other than /proc/meminfo, including /proc/mounts or /etc/mtab. I don't see how this is more secure than parsing /proc/meminfo directly in libtcmalloc.
4. If I understand you correctly, weak dependency means using `dlopen` to load libhugetlbfs dynamically. This is yet another potential security vulnerability.

alk:
On Sun, Nov 20, 2016 at 2:07 AM, Andrey Semashev <notifications@github.com>
wrote:

> Can you please elaborate on why my alternative is not acceptable for you?
>
> I already mentioned some of my concerns:
>

See below.


>    - The auto-detection code allocates a huge page, whose number is
>    limited. Another process(es) might be using most or all huge pages and the
>    auto-detection algorithm will fail. This will result in huge pages not
>    being used, even if the other process releases its pages or terminates.
>    This would be a very possible situation with our application, because there
>    are multiple (dozens) processes running, all using huge pages. Also, often
>    these processes start in packs, so it is possible that multiple processes
>    run the detection code concurrently.
>
>
We should do it only when we're about to allocate first large page anyways
and when user haven't asked for specific size. Nothing is wasted in this
case.


>    - I did not test the auto-detection code on 32-bit systems, systems
>    with 2 GiB or less memory, or systems with limited/disabled overcommit and
>    I'm not sure how the code will perform there. Even though our application
>    is currently 64-bit only, I don't want to restrict this allocator to 64-bit
>    only. Limited overcommit and low memory are supported (in the case of
>    overcommit - even the intended) use cases for our application, so the
>    allocator must perform well in these conditions.
>
>
Overcommit has nothing to do with this case due to PROT_NONE as I noted
already.

On 32-bit programs running on 64-bit machines it is likely possible to get
1 gig pages, but super-unlikely for something that large to be default. I'd
still try to temporarily reserve 2 gigs and failing that, fall back to
something smaller such as few megs. And if that fails (which is unlikely on
still very fresh address space in any 32-bit system, especially where
performance matters), just disable huge pages and require user to manually
specify size.

We can also modify our approach further to iterate size upwards until we
find size where we can actually probe the end of last huge page of our
first allocation.


>    - Lastly, I simply don't like the idea. Allocating 2 GiB of space does
>    not feel like an adequate resource usage to me.
>
>
Come on. We're temporarily allocating just a chunk of address space. Think
VirtualAlloc with just MEM_RESERVE in windows world. Not even page table
entries are allocated for PROT_NONE mapping.

Lastique:
>> - The auto-detection code allocates a huge page, whose number is
>> limited. Another process(es) might be using most or all huge pages and the
>> auto-detection algorithm will fail. This will result in huge pages not
>> being used, even if the other process releases its pages or terminates.
>> This would be a very possible situation with our application, because there
>> are multiple (dozens) processes running, all using huge pages. Also, often
>> these processes start in packs, so it is possible that multiple processes
>> run the detection code concurrently.
>>
> We should do it only when we're about to allocate first large page anyways
> and when user haven't asked for specific size. Nothing is wasted in this
> case.

It's not about wasting memory. If auto-detection fails, huge pages are 
never used by the process ever again, even if they become available.

>> - I did not test the auto-detection code on 32-bit systems, systems
>> with 2 GiB or less memory, or systems with limited/disabled overcommit and
>> I'm not sure how the code will perform there. Even though our application
>> is currently 64-bit only, I don't want to restrict this allocator to 64-bit
>> only. Limited overcommit and low memory are supported (in the case of
>> overcommit - even the intended) use cases for our application, so the
>> allocator must perform well in these conditions.
>>
> Overcommit has nothing to do with this case due to PROT_NONE as I noted
> already.
 >
> On 32-bit programs running on 64-bit machines it is likely possible to get
> 1 gig pages, but super-unlikely for something that large to be default. I'd
> still try to temporarily reserve 2 gigs and failing that, fall back to
> something smaller such as few megs. And if that fails (which is unlikely on
> still very fresh address space in any 32-bit system, especially where
> performance matters), just disable huge pages and require user to manually
> specify size.
>
> We can also modify our approach further to iterate size upwards until we
> find size where we can actually probe the end of last huge page of our
> first allocation.
 >
>> - Lastly, I simply don't like the idea. Allocating 2 GiB of space does
>> not feel like an adequate resource usage to me.
>>
> Come on. We're temporarily allocating just a chunk of address space. Think
> VirtualAlloc with just MEM_RESERVE in windows world. Not even page table
> entries are allocated for PROT_NONE mapping.

Again, I did not test it on those systems, and from the docs it doesn't 
follow that `PROT_NONE` disables page reservation. The algorithm relies 
on that fact and I don't feel comfortable relying on it. Sorry, I still 
don't like it.

alk:
Hi. I am sorry we could not reach understanding. I am rejecting this. Key thing is use of hugetlb pages requires user setup. I.e. in original message you state "Also, as I understand, using mmap directly allows to use the dynamic allocation of huge pages, when possible, while hugetlbfs only uses the preallocated pool" but this is not true.

With that in mind, people having to reserve hugetlb pages, requiring them to mount hugetlbfs is the most straightforward option. And saves all the complexities that you've proposed here. Sorry.

PR #809, opened 2016-07-01
Recognize .node files as shared libraries

kentonv:
node.js extensions are regular shared libraries named with the `.node` extension. With this small change I was able to get a usable heap profile out of node to debug an extension I'm working on.


alk:
Merged. Thanks. But do note that perl pprof implementation is deprecated and you should be working with github.com/google/pprof instead.

PR #813, opened 2016-08-02
Remove references to __malloc_initialize_hook

tuliom:
__malloc_initialize_hook has been deprecated in glibc since 2011 and is
being removed on glibc 2.24.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -456,16 +456,6 @@ AC_LINK_IFELSE([AC_LANG_PROGRAM([#if defined(__GNUC__) && ((__GNUC__ < 4) || (__
                 AC_MSG_RESULT([yes])],
                [AC_MSG_RESULT([no])])
 
-# glibc's __malloc_hook/etc were declared volatile starting in glibc 2.14
-AC_MSG_CHECKING([if __malloc_hook is declared volatile])
-AC_COMPILE_IFELSE([AC_LANG_PROGRAM([#include <malloc.h>
-void* (* volatile __malloc_hook)(size_t, const void*) = 0;],)],
-                  [AC_DEFINE(MALLOC_HOOK_MAYBE_VOLATILE, volatile,
-                             Define to 'volatile' if __malloc_hook is declared volatile)
-                   AC_MSG_RESULT([yes])],
-                  [AC_DEFINE(MALLOC_HOOK_MAYBE_VOLATILE, )
-                   AC_MSG_RESULT([no])])
-
 # Nanosleep requires extra libraries on some architectures (solaris).
 # This sets NANOSLEEP_LIBS.  nanosleep doesn't exist on mingw, which
 # is fine for us because we don't compile libspinlock, which uses it.
--- a/src/libc_override_glibc.h
+++ b/src/libc_override_glibc.h
@@ -86,40 +86,6 @@ extern "C" {
 
 #endif  // #if defined(__GNUC__) && !defined(__MACH__)
 
-
-// We also have to hook libc malloc.  While our work with weak symbols
-// should make sure libc malloc is never called in most situations, it
-// can be worked around by shared libraries with the DEEPBIND
-// environment variable set.  The below hooks libc to call our malloc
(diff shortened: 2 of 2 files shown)


alk:
Thanks for reminding me about this. But this commit isn't removing all of __malloc_hook. There are some mentions still left in override_glibc header and some in configure.ac. Would you mind dropping them too ?

tuliom:
> Would you mind dropping them too ?

I will. But do you think this is a requirement to get this patch accepted?
I'm also waiting for the conclusions of [this discussion](https://www.sourceware.org/ml/libc-alpha/2016-07/msg00662.html) to provide a final fix to statically link against glibc 2.24.

alk:
Hi again. I am not sure what is there to discuss w.r.t. statically linking libc. It currently works (I've just tested). I believe that it works because libc.a's object files that have malloc are simply not taken by linker since malloc is already taken from libtcmalloc_minimal.a (libtcmalloc.a).

As for your patch I don't have objections to merging it, but this is essentially no-op patch in this current form. So there is little value merging it.

tuliom:
I have just attached another patch to remove the rest of the glibc malloc hooks.

> It currently works (I've just tested). I believe that it works because libc.a's object files that have malloc are simply not taken by linker since malloc is already taken from libtcmalloc_minimal.a (libtcmalloc.a).

It does work with libtcmalloc_minimal.a if your program doesn't call fork, but it doesn't work with libtcmalloc.a which requires libpthread.

The current discussion is around how to avoid a patch such as tuliom/gperftools@d716579f625f5a0ae07674f019ec15f4fe802b6f

alk:
Can you please elaborate on what is not working in libtcmalloc and why forking and libpthread are relevant ?

tuliom:
[There is a patch](https://sourceware.org/git/?p=glibc.git;a=commitdiff;h=29d79486) on glibc 2.24 that added 3 new global symbols (__malloc_fork_lock_parent, __malloc_fork_unlock_parent and __malloc_fork_unlock_child) which are called by fork().  These functions are necessary on glibc malloc, but may not be necessary on other malloc implementations.
The problem is that you can't statically link libtcmalloc to libpthread 2.24 unless the 3 new symbols are overwritten as well.

The glibc community is [still discussing how to deal with this](https://sourceware.org/ml/libc-alpha/2016-07/msg00671.html).  I suggest to follow [this bug report](https://sourceware.org/bugzilla/show_bug.cgi?id=20424) in case you're interested on this topic.

Anyway, the 2 patches I have submitted here aren't related to this issue. I was just mentioning it.  ;-)

alk:
Thanks. We just do pthread atfork which we weakly depend on. So this should
still work regardless of whether pthread is linked or not.

On Aug 9, 2016 11:21 AM, "Tulio Magno Quites Machado Filho" <
notifications@github.com> wrote:

There is a patch
<https://sourceware.org/git/?p=glibc.git;a=commitdiff;h=29d79486> on glibc
2.24 that added 3 new global symbols (__malloc_fork_lock_parent,
__malloc_fork_unlock_parent and __malloc_fork_unlock_child) which are
called by fork(). These functions are necessary on glibc malloc, but may
not be necessary on other malloc implementations.
The problem is that you can't statically link libtcmalloc to libpthread
2.24 unless the 3 new symbols are overwritten as well.

The glibc community is still discussing how to deal with this
<https://sourceware.org/ml/libc-alpha/2016-07/msg00671.html>. I suggest to
follow this bug report
<https://sourceware.org/bugzilla/show_bug.cgi?id=20424> in case you're
interested on this topic.

Anyway, the 2 patches I have submitted here aren't related to this issue. I
was just mentioning it. ;-)

—
You are receiving this because you commented.

Reply to this email directly, view it on GitHub
<https://github.com/gperftools/gperftools/pull/813#issuecomment-238643873>,
or mute the thread

alk:
Applied. Thanks.

bassam:
@alk and @tuliom #856 seems related to the discussion here.

PR #829, opened 2016-09-19
Fix unaligned memory accesses in debug allocator

fjricci:
This fixes several unaligned memory accesses detected by clang's undefined behavior sanitizer.

### DIFF
--- a/src/base/basictypes.h
+++ b/src/base/basictypes.h
@@ -229,6 +229,16 @@ inline Dest bit_cast(const Source& source) {
   return dest;
 }
 
+// bit_store<Dest,Source> implements the equivalent of
+// "dest = *reinterpret_cast<Dest*>(&source)".
+//
+// This prevents undefined behavior when the dest pointer is unaligned.
+template <class Dest, class Source>
+inline void bit_store(Dest *dest, const Source *source) {
+  COMPILE_ASSERT(sizeof(Dest) == sizeof(Source), bitcasting_unequal_sizes);
+  memcpy(dest, source, sizeof(Dest));
+}
+
 #ifdef HAVE___ATTRIBUTE__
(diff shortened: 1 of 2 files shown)


alk:
Thanks for the patch. There is one problem. We cannot rely on __builtin_memcpy since it won't work on some compilers (e.g. msvc). Why we have to do it? Would it be better to ensure alignment instead?

fjricci:
So I checked the disassembly using `memcpy` instead of `__builtin_memcpy` using Clang, and they're identical (both inlined). Especially since this is in the debug allocator, I think it's reasonable to use `memcpy`, which should be portable.

This is needed because we allow `Allocate` to be called with sizes which aren't 4-byte aligned, and the checks will use these sizes as memory offsets. So, for example, if we call `Allocate` with a size of 1 byte, we will have unaligned accesses in the checks.

I'll reupload the patch using regular `memcpy`.

fjricci:
There's also a question of whether int*/size_t* is really the best return type for these functions, if we allow allocations of any size. Perhaps char* or void* would be better?

alk:
Looks good. Please try using bit_cast from basictypes.h. I am not sure, but it may help make code a bit cleaner.

fjricci:
`bit_cast` does not work, because the source and destination types are not always the same type. For example:
`*magic2_addr() = magic1_;`
`*magic2_addr()` is an `int`, but `magic1_` is a `size_t`.

So `COMPILE_ASSERT(sizeof(Dest) == sizeof(Source), bitcasting_unequal_sizes)` will fail.

alk:
Thanks. Looks like this mismatch should be fixed as well.

fjricci:
`bit_cast` doesn't quite work if the dest pointer is also unaligned. So I wrote a `bit_store` method which works similarly.

alk:
applied. Thanks.

PR #834, opened 2016-09-28
Update linux_syscall_support.h

zanderso:
The #else here defines `__syscall6` twice for `__aarch64__`. The first time is up on line 2563. Instead of a catchall for other 64-bit architectures, they can be handled explicitly.

@alk


alk:
This is failing to build with "unsupported 64-bit architecture" on my odroid c2

alk:
Merged pull request #833 instead

PR #857, opened 2016-12-17
Fix a typo in the page fence flag declaration

acmorrow:


alk:
Applied. Thanks a lot.

PR #865, opened 2017-01-19
Fix occasional crush on x86_64 environment

chen3feng:
Issue description:
With HEAPCHECK enabled environment, My program crash here occasionally.
```
// gperftools-2.4/src/stacktrace_x86-inl.h
 327       while (sp && n < max_depth) {
>328         if (*(sp+1) == reinterpret_cast<void *>(0)) {
 329           // In 64-bit code, we often see a frame that
 330           // points to itself and has a return address of 0.
 331           break;
 332         }

```   
with call stack
```
*** SIGSEGV (@0x7fff6e5999a8) received by PID 27560 (TID 0x7f2fa8fff9c0) from PID 1851365800; stack trace: ***
    @     0x7f2f94bbe710 (unknown)
    @     0x7f2f949a65d0 GetStackTrace_x86()
    @     0x7f2f949a6ba6 GetStackTrace()
    @     0x7f2f94df435a MallocHook_GetCallerStackTrace
    @     0x7f2f94e0669f HeapProfileTable::GetCallerStackTrace()
    @     0x7f2f94dfb593 NewHook()
    @     0x7f2f94df37fa MallocHook::InvokeNewHookSlow()
    @     0x7f2f9506dc39 operator new()

```
Analysis:
For performance reason, our gperftools was not configured with libunwinding,
and all our code compiled with -fno-omit-frame-pointer, but for system libraries such as glibc, we kept them unchanged, which are compiled with -fomit-frame-pointer by default. maybe it is the reason of crash.

Solution:
I noticed that msync was used in other paths to check the validity of pointer, but maybe because it is slow, It is not enabled in normal path.
So I provide an optimized implementation.

I've already signed CLA the several years ago, in the age of googlecode. so I'm glad to contribute this change.

### DIFF
--- a/src/stacktrace_x86-inl.h
+++ b/src/stacktrace_x86-inl.h
@@ -135,21 +135,6 @@ static int CountPushInstructions(const unsigned char *const addr) {
 }
 #endif
 
-// Get stack bottom of current thread
-static void* GetStackBottom() {
-  static __thread void* result;
-  if (!result) {
-    char frame;
-    uintptr_t page_size = getpagesize();
-    char* aligned_p = (char *)((uintptr_t)&frame & ~(page_size - 1));
-    while (msync(aligned_p, page_size, MS_ASYNC) != -1) {
-      aligned_p += page_size;  // x86 stack is growdown
-    }
-    result = aligned_p;
-  }
-  return result;
-}
-
 // Given a pointer to a stack frame, locate and return the calling
 // stackframe, or return NULL if no stackframe can be found. Perform sanity
 // checks (the strictness of which is controlled by the boolean parameter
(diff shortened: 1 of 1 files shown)


alk:
Hi. This is very interesting idea. But I think it won't work for signal handlers running on alternate stacks (i.e. sigaltstack).

Maybe instead of caching stack bottom, we can cache few recent results of msync?

Also I think we need to keep this code working on systems without functional __thread. config.h has define for whether TLS works, so I think you can just place ifdef around new code for TLS availability.

chen3feng:
Thanks, I'll continue work on it.

chen3feng:
Hi, I made a new change that avoided using thread local.
thank you for your review!

alk:
Thanks. How about making it even faster by caching valid stacks? Something like that: https://github.com/alk/gperftools/tree/wip-safe-and-fast-frame-pointer-unwinding but cleaned up ? https://github.com/alk/gperftools/tree/wip-safe-and-fast-frame-pointer-unwinding

Also given you're effectively dropping STRICT_UNWINDING can you please clean it up from the code ?

alk:
Hi. Do you still want this merged ?

alk:
BTW, I've now replaced x86 stacktrace capturing method with generic-fp and it now (by default) msyncs every page access. It is careful not to test same page twice. But didn't get to more advanced caching means. Anyway, this commit is now out of date. Closing.

PR #868, opened 2017-02-04
Update binary_trees.cc

jon536:


alk:
Applied. Thanks. But, please, next time send with better commit message. "update XXX" is not great commit message as it doesn't explain what is actual update.

PR #872, opened 2017-02-27
Fix a possible infinite loop in TCMalloc_ListAllProcessThreads.

yasushi-saito:
The ListerThread may access the buffer out of bounds. I have seen this thread using 100% cpu and running forever. I'm not sure if this change fixes the problem, but the original code looks suspicious.

### DIFF
--- a/src/base/linux_syscall_support.h
+++ b/src/base/linux_syscall_support.h
@@ -205,7 +205,7 @@ struct kernel_dirent64 {
   long long          d_off;
   unsigned short     d_reclen;
   unsigned char      d_type;
-  char               d_name[256];
+  char               d_name[0];
 };
 
 /* include/linux/dirent.h                                                    */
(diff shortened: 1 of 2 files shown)


yasushi-saito:
ping.

alk:
Sorry for delay. Thanks for the patch.

My understanding is the getdents won't return partial dirent. I.e. here: http://lxr.free-electrons.com/source/fs/readdir.c?v=4.9#L262

But I don't mind protecting from such possibility. I two comments on your patch itself. Can you please fix 0-sized array? While gcc accepts it, this is not valid c++ AFAIK.

Second possible issue is if we're guarding from partial dirent we should be guarding from partially filled filename. So nbytes > sizeof(struct dirent) seems not enough to protect from that.

alk:
Hi. Do you still want this merged?

yasushi-saito:
I don't care to be honest. I'm not using tcmalloc any more. Lets drop this change since nobody else seems be having the same problem.

PR #875, opened 2017-03-21
Fix path names in README

ola1olsson:


alk:
Thanks a lot. Applied.

Issue #880, opened 2017-04-21
MarkThreadTemporarilyIdle behavior may be disruptive to thread cache calculations


acmorrow:
This issue is being opened in anticipation of a PR with a proposed fix. However, we are unsure about the correctness or optimality of our proposed fix, and we would like to discuss it with the upstream maintainers. The following is background from one of the MongoDB server engineers describing what we *think* we are observing when we use MarkThreadTemporarilyIdle

> We have an application (mongod) that services incoming requests on a large number of threads. To avoid issues with contention on the central free list we call MarkThreadIdle occasionally before sleeping to wait for a new message in order to return excess memory to the central free list at a time that is unlikely to impact the thread if it is delayed due to contention. This however tears down the entire thread cache data structure so is somewhat expensive; to avoid this we created MarkThreadTemporarilyIdle which returns excess memory to the central free list without tearing down the data structures. However, counterintuitively, we observed that in some cases this actually decreased performance. In those cases we observed that thread cache free memory grew beyond the configured bounds, so we theorized that by returning memory to the central cache free list frequently we were having to then also fetch memory from the central cache free list frequently, fooling the allocator into believing that the thread was perpetually starved of locally cached memory, causing max_length_  to increase, fetching more and more memory from the central free list each time only to have to return it. To avoid this we modified MarkThreadTemporarilyIdle to reset max_length_ (by calling Init()), and this on at least one workload performed better than either calling MarkThreadIdle or calling MarkThreadTemporarilyIdle prior to our change.

CC @markbenvenuto @bdlucas1

alk:
Revisited this. So notably "abseil" version of tcmalloc is just fully destroying thread cache. So they avoid the problem. And yes problem looks real.

We should have a unit test and a fix (and likely do same as "abseil" version to keep things simpler instead of Andrew's more explicit approach).

In addition, in the commit we discussed alternative option of "pooling" hot thread caches instead of dropping objects from freelists (but failing to adjust stats). But this doesn't seem to have any demand, so just mentioning it here.

PR #881, opened 2017-04-21
Re-init free list state in BecomeTemporarilyIdle

acmorrow:
PR associated with #880.

### DIFF
--- a/src/thread_cache.cc
+++ b/src/thread_cache.cc
@@ -442,8 +442,37 @@ void ThreadCache::BecomeIdle() {
 
 void ThreadCache::BecomeTemporarilyIdle() {
   ThreadCache* heap = GetCacheIfPresent();
-  if (heap)
+  if (heap) {
     heap->Cleanup();
+
+    // Re-initialize the free list state to reset the slow-start
+    // algorithm for max_size_; this avoids requesting more and more
+    // memory from the central freelist due to frequent calls.
+    for (size_t cl = 0; cl < kNumClasses; ++cl) {
+      heap->list_[cl].Init();
+    }
+
+    // Re-calculate the Thread Cache max size if we have a non-default size.
+    if (heap->max_size_ != kMinThreadCacheSize) {
+      SpinLockHolder h(Static::pageheap_lock());
+
+      // Return claimed thread space
+      unclaimed_cache_space_ += heap->max_size_;
+
+      // Re-init max_size_
+      heap->max_size_  = 0;
+      heap->IncreaseCacheLimitLocked();
+
+      if (heap->max_size_ == 0) {
+        // There isn't enough memory to go around.  Just give the minimum to
+        // this thread.
+        heap->max_size_ = kMinThreadCacheSize;
+
+        // Take unclaimed_cache_space_ negative.
+        unclaimed_cache_space_ -= kMinThreadCacheSize;
+      }
+    }
+  }
 }
 
 void ThreadCache::DestroyThreadCache(void* ptr) {


acmorrow:
The following is additional commentary on this proposed change from another server engineer:

> While this is a nice improvement over the existing behavior, where we
> aggressively grab memory from the central free list cache after a thread is no
> longer idle, we still put a lot of burden on the CentralFreeList locks for a
> given size class. Due to max_length_ being reset to zero, we slow start our free
> list caching in ThreadCache::FetchFromCentralCache and ThreadCache::ListTooLong
> after we mark threads idle. In the first case, this means we will make a lots of
> requests to the CentralFreeList for allocations, many of the requests for sizes
> that all threads will typically make. In the second case, we will dump a lot
> memory back into the central list if we have a lot of allocations of a
> particular size followed by deallocations of the same size.
> 
> Why not factor the history of the freelist when we reset max_length_. We could
> use the previous number to choose the new one. 
> 1. We could choose a point between the previous max_length_ and 1, like 50%.
> 2. We could choose for larger pages (> 8192 or kTargetTransferBytes) to go back
> to max_length_ = 1.
> 
> As long as we choose a reasonable number where 1 < x < max_length_, we should do
> better. This would work well for cases where a particular size class sees
> similar allocation numbers between MarkTemporaryIdle calls. In the some cases, a
> thread may ask for more memory than it needs until the next MarkTemporaryIdle
> call then it needs , in that case ListToLong, and Savage will help to remove
> unused memory.
> 
> Thoughts?
> 
> ```
>     // After going idle, we need to reset the low watermark to control
> scavenging,
>     // and reset the max_length_ so we fetch reasonable number of objects from
> the
>     // central cache on the first allocation for a given class. For max_length_,
> 
>     // use the previous max_length / 2 as a starting point, but limit this to a 
>     // batch_size at the most.
>     void ReInitFromIdle(size_t cl) {
>       const int batch_size = Static::sizemap()->num_objects_to_move(cl);
>       ASSERT(list_->empty());
>       list_ = NULL;
>       length_ = 0;
>       lowater_ = 0;
>       max_length_ = min<int>((max_length_ + 1) / 2, batch_size);
>       // Use slow start for larger size classes.
>       if (Static::sizemap()->ByteSizeForClass(cl) > 8192)
>           max_length_ = 1;
>       length_overages_ = 0;
>     }
> ```

CC @markbenvenuto @bdlucas1

acmorrow:
@alk We would appreciate your thoughts on this issue.

alk:
The idea looks reasonable. How about alternative proposal. Say when thread goes idle or dies, it's entire thread cache is not destroyed but goes into pool instead. Then next thread becoming active can grab that "hot" thread cache and be fast.

I think it is fundamentally much more friendly to "thread per connection" model that your server seems to be doing. Let me know what you think.

alk:
Looked at the patch. I don't mind merging it, but IMHO a bit more generic approach, as I've proposed above is possible and should work better. Let me know what you think. Alternatively, if there are any single node tests I can run on my reasonably sized laptop, let me know as well.

alk:
I thought some more about it. Looks like re-setting free lists without adjusting unclaimed_cache_space_ is going to "leak" space in thread cache total size.

And I still think our best bet to support thread per connection mode, from complexity perspective and from performance perspective, is to do "idle" thread cache pooling. Of course, per-cpu cache would work even better, but they require kernel support which won't happen retrospectively for stuff like rhel6 or rhel7, even if it can be accepted to upstream kernel.

acmorrow:
@alk Thanks for your comments. We are discussing internally. I do have a few quick comments:

- We are *definitely* interested in a per-cpu mode for tcmalloc, more like the Solaris libumem allocator. It is something we have talked about a fair amount, and I even started trying it out with tcmalloc once, but got distracted.
- If you could provide a little more color on your comment about 'leak space in thread cache total size' it would be appreciated.
- We are not completely sure that the proposal of pooling entire thread caches would allow us to achieve the (intended) objective of MarkThreadTemporarilyIdle to overall decrease the total amount of heap held cached in thread caches, since you would still have pooled caches, even if they weren't currently associated with any thread.

We are planning to re-visit some of our original perf studies that lead to the decisions around MarkThreadTemporarilyIdle so we can continue this discussion with more data. Looking forward to your thoughts on the above and we will let you know when we have more information to share as well.

alk:
per-cpu caches: so I am not sure what libumem does. If you can send me some links to the code, I'd be curious reading it. What I am referring to as per-cpu mode is based on restartable sequences support in kernel (https://lwn.net/Articles/650333) which allows implementing per-cpu caches without doing any expensive atomic operations. We're talking about couple dozen cheap instructions per fast-path of malloc or free. Without kernel support, which won't ever be backported to e.g. rhel 6, we'll need to keep per-thread mode competitive.

As for leaking, I've incorrectly confused max_length_ that is being reset with max_size that isn't.

As for pooling. Here is my understanding of the problem. We have fixed memory limit for total thread caches size. When there are plenty of at least somewhat active thread caches, they're competing for thread cache size and doing too much of slow path (central free list stuff). The issue is there are too many thread caches if we're doing thread per connection. And efficiency is low.

With pooling, we would only have thread caches for actually active threads. Which will likely be limited to small tens under even quite io-heavy load. So with same total thread cache size there won't be as much churn w.r.t. thread cache sizes etc. And this is simpler to implement and reason about than all this mark idle logic. IMHO.

markbenvenuto:
You are right that `unclaimed_cache_space_` is leaking memory in our proposed patch. It is a fix we need to make to the proposed change.

On MongoDB's threading model, we do have a thread per connection in our released versions (which we are working to change for the next major version), and as a result see high-contention on the SpinLocks in the slow path of CentralFreeList. 

In our workload, we often see asymmetric memory allocation in most scenarios. On any given thread, we may read database pages on one thread but evict them on a different thread. Or, clients may issue different kinds of requests on the same thread which create different memory allocation workloads from tcmalloc's perspective.

The "idle" thread hack on our connection threads chiefly helps to mitigate the lock contention we see in `ThreadCache::ListToLong()`. We see the contention on our longer test runs over time as `max_length_` increases in length, and larger, and larger lists of items are pushed back to the `ThreadCache::ReleaseToCentralCache`. Most of our released supported versions are running with 2.2 which has `kDefaultTransferNumObjecs = 32768`. I plan to experiment with undoing our "idle" hack, and seeing how running with `kDefaultTransferNumObjecs = 32` affects performance from your recent changes.

acmorrow:
@alk and @markbenvenuto - I've updated this PR to reflect MongoDB's current thoughts on how to address the issues raised in the above discussion. We would love to hear your feedback.

alk:
Thanks for the patch. BTW it is a bit out-of-date w.r.t. state of master. This is patch to update it:

diff --git a/src/thread_cache.cc b/src/thread_cache.cc
index f0ffa66..987b01a 100644
--- a/src/thread_cache.cc
+++ b/src/thread_cache.cc
@@ -448,8 +448,8 @@ void ThreadCache::BecomeTemporarilyIdle() {
     // Re-initialize the free list state to reset the slow-start
     // algorithm for max_size_; this avoids requesting more and more
     // memory from the central freelist due to frequent calls.
-    for (size_t cl = 0; cl < kNumClasses; ++cl) {
-      heap->list_[cl].Init();
+    for (size_t cl = 0; cl < Static::num_size_classes(); ++cl) {
+      heap->list_[cl].Init(Static::sizemap()->class_to_size(cl));
     }
 
     // Re-calculate the Thread Cache max size if we have a non-default size.


As for the patch itself, I am not sure I get what  is the thinking here. Particularly w.r.t. difference with BecomeIdle. The only difference I see is manipulation of unclaimed space and max_size. And it looks counter-intuitive. We're about to go to sleep for potentially long long time, yet we're reserving ourselves part of global cache size quota. Doing ~same at wake up time would seem to make a bit more sense.

acmorrow:
Thanks @alk for taking a look. @markbenvenuto could you please get back to @alk w.r.t. his comments above, and make the necessary updates for master?

alk:
BTW as part of malloc tracing work I need facility similar to what I've described above.

See this commit https://github.com/alk/gperftools/commit/f452eaa0fef6ed4da8d0f5a7bb6804dfb86f4a7c for {release,set}_malloc_thread_cache API.

The idea is when thread goes idle, it'll extract it's thread cache and put it into some pool. Then when (potentially other thread) wakes up from idleness, it grabs "hot" thread cache from pool and things are still fast.

There will likely have to be more extensions w.r.t. managing this idle pool. Say, would be nice if we could "gc" caches of this pool when we need thread cache size quota either/or more ram (say just before growing heap we scavenge potentially cold caches).

I have very high confidence that approach like that is going to work quite well for programs that have lots of idle threads. E.g. using thread per connection model like mongo.

I don't yet plan to add this API to master branch. I am only using it for quick trace replaying for now. But I am very interested in your opinion on whether we should have production api like that.

Of course another possible answer to efficient thread-per-connection handling is per-cpu caches. But those still need restartable sequences or similar to be really efficient and they have their own problems on shared machines with lots of cpus.

alk:
You guys still want this? I think your program would benefit from being able to have idle thread quickly return it's entire thread cache to common pool of thread caches, so that whenever other thread un-becomes idle, it can grab hot and ready cache from there.

This is direction I would consider. And I think it could even beat performance of per-cpu caches mode, if we do it right.

acmorrow:
@alk - It is something of a tough call. I think the feature has merit. However, I also managed to get MongoDB to stop using idle marking entirely in a recent release. So we are no longer using it. I think we can probably close it and if it becomes interesting again in the future we can always re-open, or revisit under the new project.

alk:
Revisited this. So notably "abseil" version of tcmalloc is just fully destroying thread cache. So they avoid the problem. And yes problem is real.

Anyways, dropping this commit as discussed. But I'll keep original ticket open.

PR #883, opened 2017-04-26
Fix compilation error for powerpc32

RajalakshmiS:
Fix the following compilation error for powerpc32 platform when using
latest glibc.
error: ‘siginfo_t’ was not declared in this scope

### DIFF
--- a/src/stacktrace_powerpc-linux-inl.h
+++ b/src/stacktrace_powerpc-linux-inl.h
@@ -44,6 +44,7 @@
 
 #include <stdint.h>   // for uintptr_t
 #include <stdlib.h>   // for NULL
+#include <signal.h>  // for siginfo_t
 #include <gperftools/stacktrace.h>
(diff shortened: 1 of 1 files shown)


zatrazz:
I think we can just include signal.h for all powerpc abis.

RajalakshmiS:
@zatrazz At this point, it is needed only for ppc32, If needed we can include it for other abis in future.

alk:
Yes, let's please avoid unnecessary ifdef and include unconditionally.

RajalakshmiS:
@alk @zatrazz Updated as per comment.

alk:
Can you please squash those 2 changes into 1 commit ? Then I'll happily merge it.

RajalakshmiS:
@alk Can you please check if the latest commit is fine?

alk:
Applied. Thanks.

PR #885, opened 2017-05-11
remove superfluous size_t value >= 0 check

mostynb:
This assertion adds no value, and can trigger warnings (and errors if you build with -Werror).  I think it should be removed.  If it is really important, a comment where the variable is defined would suffice.


alk:
Merged. Thanks.

PR #893, opened 2017-05-26
Document HEAPPROFILESIGNAL environment variable

kimgr:
We've found this relatively new config option highly useful. It would be nice if it was included in the user-facing documentation.


alk:
Applied. Thanks.

kimgr:
Thanks!

PR #896, opened 2017-06-15
Ignore current_instance heap allocation when leak sanitizer is enabled

fjricci:
Without this patch, any user program that enables LeakSanitizer will
see a leak from tcmalloc. Add a weak hook to __lsan_ignore_object,
so that if LeakSanitizer is enabled, the allocation can be ignored.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -79,6 +79,9 @@ TCMALLOC_FLAGS =
 if MINGW
 TCMALLOC_FLAGS += -Wl,-u__tcmalloc
 endif MINGW
+if OSX
+AM_LDFLAGS += -Wl,-U,___lsan_ignore_object
+endif OSX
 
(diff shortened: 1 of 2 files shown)


alk:
Thanks a lot for the patch. Can you add some comments on why -U flag is needed on osx but no anywhere else? (I couldn't find it in man page of ld on my GNU/Linux system)

alk:
Found this: http://www.manpages.info/macosx/ld.1.html so seemingly makes sense. I don't think you need to bother with adding AM_LDFLAGS to libtcmalloc_la_LDFLAGS. I think AM_LDFLAGS is added automagically.

fjricci:
Yeah, osx errors on undefined symbols by default, so -U is needed when you use weak hooks.

alk:
Applied. Thanks.

PR #897, opened 2017-06-18
Fix c++17 build

Romain-Geissler-1A:
Hi,

C++17 deprecated the usage of "throw(std::bad_alloc)" which causes error when building with gcc 7 and -std=gnu++17.

Here is a small patch set to fix that:
 - Rename PERFTOOLS_THROW into PERFTOOLS_NOTHROW (less counter intuitive name)
 - Add PERFTOOLS_THROW where necessary (as detected by GCC)
 - Replace "throw()" by "PERFTOOLS_NOTHROW" (not necessary, but I find it cleaner to always use the same syntax IMO).

Cheers,
Romain

### DIFF
--- a/src/base/basictypes.h
+++ b/src/base/basictypes.h
@@ -433,4 +433,11 @@ namespace base {
 enum LinkerInitialized { LINKER_INITIALIZED };
 }
 
+#if __cpp_noexcept_function_type >= 201510
+// Deprecated in C++17
+# define PERFTOOLS_THROW(...)
+#else
+# define PERFTOOLS_THROW(...) throw(__VA_ARGS__)
+#endif
+
 #endif  // _BASICTYPES_H_
--- a/src/debugallocation.cc
+++ b/src/debugallocation.cc
@@ -1227,7 +1227,7 @@ static void force_frame() {
   (void)dummy;
 }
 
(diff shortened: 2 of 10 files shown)


alk:
Thanks for the patch. Can you please squash latest commit into second and move PERFTOOLS_THROW out of public header ? Other than that, your patch looks perfect.

Romain-Geissler-1A:
I have moved PERFTOOLS_THROW out of public header in a new commit. Is this OK for you or do you want to merge that with earlier commits ?

alk:
Yes, I prefer clean history.

Romain-Geissler-1A:
Done

alk:
Hm. Somehow it broke compilation on visual studio and mingw. Looks like win32 port.h is doing #define __THROW PERFTOOLS_NOTHROW and somehow either necessary include is missing or something else is doesn't work.

Romain-Geissler-1A:
I have reverted the change in src/windows/port.h now.

I am unable to test under windows, so please tell me if it's fine now.

alk:
Applied. Thanks. I made slight change in your last commit (and updated message accordingly) to only define empty XXX_NOTHROW if standard is less than c++11. Less confusing this way.

PR #902, opened 2017-07-05
Ensure that lsan flags are appended on all necessary targets

fjricci:


alk:
Applied. Thanks a lot and apologies for troubles.

PR #906, opened 2017-07-10
Generalized --with-tcmalloc-pagesize=N

yurivict:
Now it accepts any power of 2: 1, 2, 4, 8, 32, 64, ...
Setting higher values may deliver a performance benefit for the expense of memory consumption.

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -113,8 +113,9 @@ To build libtcmalloc with large pages you need to use the
 
    ./configure <other flags> --with-tcmalloc-pagesize=32
 
-The ARG argument can be 8, 32 or 64 which sets the internal page size to
-8K, 32K and 64K repectively. The default is 8K.
+The ARG argument can be 1, 2, 4, 8, 16, 32, 64, ..., which sets the
+internal page size to 1K, 2K, 4K, 8K, 32K, 64K, ... respectively.
+The default is 8K.
 
 
 *** SMALL TCMALLOC CACHES: TRADING SPACE FOR TIME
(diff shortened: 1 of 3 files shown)


alk:
Thanks a lot for the patch. I don't think it'll work as is. The issue is span's field for number of free objects is 16-bit: https://github.com/gperftools/gperftools/blob/e033431e5a9e02e4d1882bb6fcd212c4fab56173/src/span.h#L51

There may be other limitations that I am not seeing yet.

Also we cannot afford advanced shell script in configure. I think it makes sense to simply list all valid options which should be quite limited set.

yurivict:
Ok. Then please do it yourself. -)

yurivict:
@alk

Aliaksey, I changed the syntax to POSIX shell (or Bourne shell).
Now there is no advanced shelling at all. -)

Cheers,
Yuri

yurivict:
I didn't notice your previous message.
If to list them, this will be 1,2,4,8,16,32,64? (K)

alk:
We can go up to 256k I think. There are some lenth/size types that are 16 bit. So for 8 bytes size class, we'll be capped at 256k.

There are actually programs at google that benefit from 256k malloc page size. So supporting it in gperftools is nice to have.

muscleprime:
Hi,
How can I get the page size of an already installed gperftools?
Thanks

alk:
>> How can I get the page size of an already installed gperftools?

There seem to be no API for that. General thinking is malloc's logical page size is implementation detail.

alk:
Just merged alternative implementation from mongo folks. --with-tcmalloc-pagesize=128 (128 is 128k actually) now works \o/

So dropping your patch.

PR #907, opened 2017-07-14
implement GetProgramInvocationName on FreeBSD.

movsw:
Code from /usr/src/contrib/libexecinfo/backtrace.c


alk:
Thanks a lot. This looks right. Can you please check if other BSDs support this as well? Based on manpage it looks like there is good chance.

Also please fix indentation on "return program_invocation_name" line. And I'd strongly prefer commit "subject" line to be shorter and saying more specifically what was done. I'd propose "implement GetProgramInvocationName on BSD" or something like that.

movsw:
Unfortunately, I only have FreeBSD at hand.

The indentation has been fixed and subject has been renamed according to your recommendation.

alk:
Applied. Thanks.

PR #908, opened 2017-07-16
Use ucontext_t instead of struct ucontext

kraj:
Newer glibc has dropped the ucontext tag from exposing

Signed-off-by: Khem Raj <raj.khem@gmail.com>


alk:
Thanks a lot. Applied.

PR #914, opened 2017-08-01
Some minor bug fix

cs-lee:
1. va_end() missing - src/symbolize.cc - SymbolTable::Symbolize
2. memory leak - src/debugallocation.cc - TracePrintf


cs-lee:
@alk Please review this patch. Thank you.

alk:
Thanks. Those two fixes look right. Can you please send them as two commits? Also please make sure that first line of each commit message describes what is fixed.

cs-lee:
ok I will separate this and request again.

PR #915, opened 2017-08-06
Some minor bug fixes

cs-lee:
va_end() missing in TracePrintf function
memory leak fix in Symbolize function


cs-lee:
@alk I have separated the commit. Please review. Thank you.

alk:
Thanks a lot. Applied.

PR #927, opened 2017-09-21
Fixed LTO warning for return values for ProfilingIsEnabledForAllThreads

dorinlazar:
Fixed LTO warning about the mismatch between return values for ProfilingIsEnabledForAllThreads().

When compiled with -flto and linking against a library compiled as such I get a warning about mismatched return type. Attached is the fix.


alk:
Thanks a lot. Applied.

PR #928, opened 2017-09-30
Add support for C++17 operator new/delete for overaligned types

Lastique:
- Add auto-detection of `std::align_val_t` presence to configure scripts. This
  indicates that the compiler supports C++17 operator new/delete overloads
  for overaligned types.

- Add auto-detection of `-faligned-new` compiler option that appeared in gcc 7.
  The option allows the compiler to generate calls to the new operators. It is
  needed for tests.

- Added overrides for the new operators. The overrides are enabled if the
  support for `std::align_val_t` has been detected. The implementation is mostly
  based on the infrastructure used by `memalign`, which had to be extended to
  support being used by C++ operators in addition to C functions. In particular,
  the debug version of the library has to distinguish memory allocated by
  `memalign` from that by `operator new`. The current implementation of sized
  overaligned `delete` operators do not make use of the supplied size argument
  except for the debug allocator because it is difficult to calculate the exact
  allocation size that was used to allocate memory with alignment. This can be
  done in the future.

- Removed forward declaration of `std::nothrow_t`. This was not portable as
  the standard library is not required to provide `nothrow_t` directly in
  `namespace std` (it could use e.g. an inline namespace within `std`). The `<new>`
  header needs to be included for `std::align_val_t` anyway.

- Fixed `operator delete[]` implementation in `libc_override_redefine.h`.

- Moved `TC_ALIAS` definition to the beginning of the file in `tcmalloc.cc` so that
  the macro is defined before its first use in `nallocx`.

- Added tests to verify the added operators.

I only tested this on Linux with gcc 7.

The relevant C++17 proposal is [P0035R4](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2016/p0035r4.html).

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -51,6 +51,9 @@ endif HAVE_W_NO_UNUSED_RESULT
 if HAVE_SIZED_DEALLOCATION
 AM_CXXFLAGS += -fsized-deallocation
 endif HAVE_SIZED_DEALLOCATION
+if HAVE_F_ALIGNED_NEW
+AM_CXXFLAGS += -faligned-new
+endif HAVE_F_ALIGNED_NEW
 
 # The -no-undefined flag allows libtool to generate shared libraries for
 # Cygwin and MinGW.  LIBSTDCXX_LA_LINKER_FLAG is used to fix a Solaris bug.
--- a/configure.ac
+++ b/configure.ac
@@ -382,6 +382,44 @@ AC_CACHE_CHECK([if C++ compiler supports -fsized-deallocation],
 AM_CONDITIONAL(HAVE_SIZED_DEALLOCATION,
                test "$perftools_cv_sized_deallocation_result" = yes)
 
+AC_CACHE_CHECK([if C++ compiler supports std::align_val_t without options],
+               [perftools_cv_have_align_val_t],
+               [AC_LANG_PUSH(C++)
+                AC_COMPILE_IFELSE([AC_LANG_PROGRAM(
+                    [[#include <new>]],
+                    [[(::operator delete)((::operator new)(256, std::align_val_t(16)), std::align_val_t(16))]])],
+                 perftools_cv_have_align_val_t=yes,
+                 perftools_cv_have_align_val_t=no)
+                AC_LANG_POP(C++)])
+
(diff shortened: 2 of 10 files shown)


alk:
Thanks very much. On first sight it looks plausible. I'll take closer look tomorrow.

One thing I spotted is ifunc logic for aligned delete seems unnecessary. There is long story of me trying to make ifunc work, but it there is just no safe way to inspect environment when ifuncs seem to run. More so when eager symbol resolution is enabled (which i.e. redhats do for security reasons).

Lastique:
> One thing I spotted is ifunc logic for aligned delete seems unnecessary.

I followed the existing code. I figured, that branch is for a platform with a non-gcc-compatible compiler but with GNU loader (not that I know such platform). Let me know if you want me remove that code.

alk:
Ah, indeed, you're duplicating "dynamic sized delete" logic. This is fine.

As for why it is needed. The idea was let distros ship libtcmalloc.so with support for sized deallocation but have it disabled by default (because lots of code out there isn't conforming to sized delete). And let either users switch it on or have programs (e.g. mongodb) enable it at run time.

It doesn't work because ifunc and getenv aren't playing along, sadly. Particularly when eager relocations are set up.

I'll need another pass over this code. Expect me to do it sometime soon.

alk:
One thing that I noticed so far is that we're aliasing functions with different signatures. While this works in practice on seemingly all architectures of today, it seems wrong. Lets please fix that.

I know that we already "incorrectly" alias tc_delete and tc_delete_nothrow. I'll be fixing it separately.

Lastique:
Sorry for the delay.

> One thing that I noticed so far is that we're aliasing functions with different signatures. While this works in practice on seemingly all architectures of today, it seems wrong. Lets please fix that.

Done.

alk:
Thanks. Applied with some minor updates. Also decided to reintroduce aliasing for performance reasons. The later is done as separate commit. Feel free to review when you have time.

alk:
Ah, another notable thing is if/when you have time, please consider adding debugallocation checks that aligned objects are freed properly (e.g. with right alignment).

PR #930, opened 2017-10-15
Revert the size_left_ to size_ in ThreadCache to fix the data race in the ThreadCache::SetMaxSize

Jianbo-Yang:
1.  Delete size_left_ and add size_
2.  Modify ThreadCache::SetMaxSize
3.  Change the logic of ThreadCache::ListTooLong introduced by the PR with size_left_
4.  Change the logic of ThreadCache::DeAllocate accordingly

### DIFF
--- a/src/thread_cache.h
+++ b/src/thread_cache.h
@@ -385,19 +385,26 @@ inline ATTRIBUTE_ALWAYS_INLINE void ThreadCache::Deallocate(void* ptr, uint32 cl
   ASSERT(list_[cl].max_length() > 0);
   FreeList* list = &list_[cl];
   size_ += Static::sizemap()->ByteSizeForClass(cl);
+  ssize_t size_headroom = max_size_ - size_ - 1;
 
   // This catches back-to-back frees of allocs in the same size
   // class. A more comprehensive (and expensive) test would be to walk
   // the entire freelist. But this might be enough to find some bugs.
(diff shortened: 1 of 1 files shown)


alk:
Thanks a lot for the patch. This looks roughly right. Just few things:

*) there seem to be no value from "bit trick" anymore.

*) Tail-calling Scavenge or ListTooLong is actually critical for performance. You can see by disasming tc_delete (or tc_delete_sized) before/after. Tail call helps compiler avoid pushing any registers on stack which is important savings for code that is as hot as malloc fast path. So please return Scavenge logic into ListTooLong.

*) it is slightly better to take object size from thread cache's freelist instead of going to sizemap as your commit does

*) I see no value splitting this into 3 commits. Lets have just one commit. Also would be really great if you follow common commit message style (e.g: https://chris.beams.io/posts/git-commit/, except I instead of 50 chars per subject line I use 65). You can google "commit message style" for some more pointers.

I think you'll get roughly same performance numbers from ./malloc_bench after you fix tail call (I do prefer object size thing too; it seems less important for fast-path performance on x86 but might matter for other achitectures)

Jianbo-Yang:
@alk sorry about the bad commit style and thank you for pointing those out.
I saw the significance of tail-calling from your previous commit which introduced size_left_. I will utilize this.
And I will also fix other issues you mentioned, then make a new pull request.
Thank you again.

PR #931, opened 2017-10-16
Fix data race setting size_left_ in ThreadCache::SetMaxSize

Jianbo-Yang:
This commit is to fix the data race in ThreadCache::SetMaxSize.
ThreadCache::size_left_ is removed and ThreadCache::size_ is
added. ThreadCache::size_left_ was introduced for optimization.
It is updated in several functions of ThreadCache, including the
ThreadCache::SetMaxSize. But thread A can update size_left_ of
thread B via SetMaxSize without protection or synchronization.

There should not be data race around ThreadCache::size_, for it
isn't accessed by multi threads.

The optimization of tail-call in tc_{malloc, new, free} is kept
and no other logics are affected.

### DIFF
--- a/src/thread_cache.cc
+++ b/src/thread_cache.cc
@@ -76,7 +76,7 @@ bool ThreadCache::tsd_inited_ = false;
 pthread_key_t ThreadCache::heap_key_;
 
 void ThreadCache::Init(pthread_t tid) {
-  size_left_ = 0;
+  size_ = 0;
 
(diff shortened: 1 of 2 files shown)


Jianbo-Yang:
@alk I made a new PR to fix this issue, please review and give comments. Thank you so much

alk:
This is great. Applied. Huge thanks.

PR #938, opened 2017-11-17
CMake build

isaachier:
This is very much a work in progress. Close to done, but I am having some issues with cyclic dependencies in the libraries. Any thoughts on separating some of the stacktrace stuff out of the tcmalloc archive?


alk:
Thanks for patch. I am not best person to review this since I deliberately choose to remain ignorant w.r.t. all things CMake. I could apply it as-is.

As of now it fails build on my machine.

isaachier:
Can I see the output? No guarantees about completing this seeing how long it has been since I opened this.

alk:
https://pastebin.com/Q7VaybKY

Qix-:
@alk What compiler are you using?

```c
#if HAVE_STRUCT_MALLINFO
```

I assume `HAVE_STRUCT_MALLINFO` is not defined, which (strangely) is making your compiler error with `#if without expression`, which is objectively not true. Seems... wrong.

:+1: to cmake configuration.

CMakeLists.txt
```diff
@@ -0,0 +1,857 @@
...
+project(gperftools VERSION 2.6.1 LANGUAGES C CXX)
+
+set(tcmalloc_so_version 8.5.4)
+set(profiler_so_version 4.14.4)
+
+set(default_build_cpu_profiler ON)
+set(default_build_heap_profiler ON)
+set(default_build_heap_checker ON)
+set(default_build_debugalloc ON)
+set(default_build_minimal OFF)
```
Qix-: CMake conventions dictate using upper case for all variable names.

CMakeLists.txt
```diff
@@ -0,0 +1,857 @@
...
+include(CheckCSourceCompiles)
+include(CheckCXXSourceCompiles)
+include(CheckFunctionExists)
+include(CheckIncludeFile)
+include(CheckLibraryExists)
+include(CheckSymbolExists)
+include(CheckTypeSize)
+include(CheckVariableExists)
+include(CMakeDependentOption)
+include(CTest)
```
Qix-: AFAIK `CTest` is included automatically. `add_test()` is a builtin - unless you're using it for something else?

CMakeLists.txt
```diff
@@ -0,0 +1,857 @@
...
+if(PPC64)
+  set(default_enable_libunwind OFF)
+  set(default_tcmalloc_pagesize 64)
+else()
+  set(default_enable_libunwind ON)
+  set(default_tcmalloc_pagesize 8)
+endif()
+
+option(
+  gperftools_build_minimal
```
Qix-: Same thing here - upper case is the best practice.

CMakeLists.txt
```diff
@@ -0,0 +1,857 @@
...
+    set(CMAKE_REQUIRED_LIBRARIES execinfo)
+    check_function_exists("backtrace" backtrace_exists)
+    set(CMAKE_REQUIRED_LIBRARIES)
+    if(backtrace_exists)
+      list(INSERT unwind_libs 0 execinfo)
+    endif()
+  endif()
+endif()
+
+find_package(Threads REQUIRED)
```
Qix-: Are threads _actually_ required? I recall being able to disable threading for tcalloc.

CMakeLists.txt
```diff
@@ -0,0 +1,857 @@
...
+if(objcopy_exe)
+  set(weaken
+      -W malloc -W free -W realloc -W calloc -W cfree
+      -W memalign -W posix_memalign -W valloc -W pvalloc
+      -W aligned_alloc
+      -W malloc_stats -W mallopt -W mallinfo -W nallocx
+      -W _Znwm -W _ZnwmRKSt9nothrow_t -W _Znam -W _ZnamRKSt9nothrow_t
+      -W _ZdlPv -W _ZdaPv
+      -W __Znwm -W __ZnwmRKSt9nothrow_t -W __Znam -W __ZnamRKSt9nothrow_t
+      -W __ZdlPv -W __ZdaPv)
```
Qix-: This variable doesn't appear to be used anywhere. I was checking usage because this will initialize `weaken` with `-W:malloc:-W:free:-W:realloc` instead of spaces. Is this intentional? It's uncommon to use flags outside of double quotes.

CMakeLists.txt
isaachier: Sure

CMakeLists.txt
isaachier: `BUILD_TESTING` is only defined when including this module. See https://cmake.org/cmake/help/latest/module/CTest.html.

CMakeLists.txt
isaachier: Let me look into this.

CMakeLists.txt
isaachier: Double quotes are only necessary for interpolation and potentially ambiguous phrases. In CMake, pretty much everything defaults to a string. If you'd like, I can change it.

CMakeLists.txt
Qix-: That's by design. `BUILD_TESTING` is off by default, for good reason. Most users define `-DBUILD_TESTING=On` in order to automatically build tests. Then the workflow becomes simply `make` and `ctest` to run the entire project's suite of tests.

CMakeLists.txt
Qix-: That's not entirely true, though. `set()` with multiple arguments (beyond the initial name/type) create a list.

```cmake
set (FOO_BAR a b c d e f)
set (FOO_BAR2 "a b c d e f")
message (STATUS "${FOO_BAR}")
message (STATUS "${FOO_BAR2}")
```

outputs

```
-- a;b;c;d;e;f
-- a b c d e f
```

isaachier:
> @alk What compiler are you using?
> 
> ```c
> #if HAVE_STRUCT_MALLINFO
> ```
> 
> I assume `HAVE_STRUCT_MALLINFO` is not defined, which (strangely) is making your compiler error with `#if without expression`, which is objectively not true. Seems... wrong.
> 
> +1 to cmake configuration.

I think this is indeed an error. I define use `#define HAVE_STRUCT_MALLINFO` but it has no associated value. Either I need to set it to 1 or use `#ifdef` in the header instead of `#if`.

Qix-:
Right, just checked - you're right @isaachier.

```c
#include <stdio.h>

#ifdef FOOBAR
#undef FOOBAR
#endif

int main(void) {
#if FOOBAR
	puts("FOOBAR in first");
#endif
#define FOOBAR
#if FOOBAR
	puts("FOOBAR in second");
#endif
#undef FOOBAR
#define FOOBAR 1
#if FOOBAR
	puts("FOOBAR in third");
#endif
	return 0;
}
```

Errors with the same message @alk got.

CMakeLists.txt
isaachier: Ya so this is one of the worse aspects of CMake. Lists are strings too, just delimited by semicolons. The problem with using a large string that contains all the flags is that COMMAND wants a list of args, not a string.

Qix-: Yeah :| As long as it's not used within a string interpolation expression, they'll work as expected.

I agree CMake misses the mark here.

Bruce1634:
i have cloned the code, and build, but only .a files apppear, if is it normal?
i compile by:
mkdir build && cd build && cmake ../
make -j8

i can not find *.so

and make install is also not working.

how can i use cmake to make *.so?
thx very much!!!!

alk:
Hi all. This still fails to build on my fairly standard debian system. And as pointed out above, it only builds .a libraries. At some point I'll face a choice of either closing this pull request or merging it's suboptimal form (assuming it at least builds on modern GNU/Linux).

okhowang:
@isaachier any progress now?
If I want to add CMake support to gperftools.
Do I have to open a new PR?

alk:
Thanks a lot for this effort. I merged it via pull #1214. So closing this one.

Issue #886, opened 2017-05-15
tcmalloc micro-benchmarks are pathetically slow on OSX


alk:
Running malloc_bench microbenchmarks on OSX I see that they're about 5 times slower than on GNU/Linux. It seems to be still about 50% faster then osx own malloc.

Part of it is due to lack of performant thread-local storage support, but there are also layers of osx own malloc that has to be run before mz_malloc is called that we're overriding.

If someone cares about tcmalloc performance on osx, this seems like something we could fix.

alk:
Also benchmarking tc_malloc/tc_free directly gives around 11 nanoseconds per loop (still about 2x slower than gnu/linux, apparently due to tls), but benchmarking malloc/free or operator new/operator delete is around 30 ns.

alk:
it is actually less pathetic now (when operator new/delete is used). Lets assume this closed. If someone steps up and contributes more improvements, fine, but no need to keep this ticket.

PR #943, opened 2017-12-05
configure.ac: better test for -faligned-new

### LINKED ISSUE #942: gpertools 2.6.2 fails to compile with Mac OS Sierra/XCode 9.2
Here's my [config.log](https://github.com/gperftools/gperftools/files/1529200/config.log).

`make install` fails with:

```
libtool: link: g++ -D_THREAD_SAFE -Wall -Wwrite-strings -Woverloaded-virtual -Wno-sign-compare -fno-builtin-malloc -fno-builtin-free -fno-builtin-realloc -fno-builtin-calloc -fno-builtin-cfree -fno-builtin-memalign -fno-builtin-posix_memalign -fno-builtin-valloc -fno-builtin-pvalloc -Wno-unused-result -fsized-deallocation -faligned-new -fno-builtin -g -O2 -D_THREAD_SAFE -o .libs/tcmalloc_unittest src/tests/tcmalloc_unittest-tcmalloc_unittest.o src/tests/tcmalloc_unittest-testutil.o -Wl,-bind_at_load  ./.libs/libtcmalloc.dylib ./.libs/liblogging.a
Undefined symbols for architecture x86_64:
  "operator delete[](void*, unsigned long, std::align_val_t)", referenced from:
      testing::RunAllTests(int, char**) in tcmalloc_unittest-tcmalloc_unittest.o
  "operator delete(void*, std::align_val_t)", referenced from:
      testing::RunAllTests(int, char**) in tcmalloc_unittest-tcmalloc_unittest.o
  "operator delete(void*, std::align_val_t, std::nothrow_t const&)", referenced from:
      testing::RunAllTests(int, char**) in tcmalloc_unittest-tcmalloc_unittest.o
  "operator delete(void*, unsigned long, std::align_val_t)", referenced from:
      testing::RunAllTests(int, char**) in tcmalloc_unittest-tcmalloc_unittest.o
  "operator new[](unsigned long, std::align_val_t)", referenced from:
      testing::RunAllTests(int, char**) in tcmalloc_unittest-tcmalloc_unittest.o
  "operator new[](unsigned long, std::align_val_t, std::nothrow_t const&)", referenced from:
      testing::RunAllTests(int, char**) in tcmalloc_unittest-tcmalloc_unittest.o
  "operator new(unsigned long, std::align_val_t)", referenced from:
      testing::RunAllTests(int, char**) in tcmalloc_unittest-tcmalloc_unittest.o
  "operator new(unsigned long, std::align_val_t, std::nothrow_t const&)", referenced from:
      testing::RunAllTests(int, char**) in tcmalloc_unittest-tcmalloc_unittest.o
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make: *** [tcmalloc_unittest] Error 1
```

I think what's happening is that the newest versions of xcode (I saw this error with 9.1 and 9.2) support `-faligned-new` which enables both aligned and size new/delete operators. 

I believe this can be fixed by providing prototypes for the `tc_new_*` and `tc_delete_*` variants in   `gperftools/tcmalloc.h` when `ENABLE_ALIGNED_NEW_DELETE` is defined and by adding definitions for the aligned/sized new/delete to `libc_override_osx.h`. I hope to publish a PR with a fix shortly.

zuercher:
Apple's clang enables aligned new/delete and sized deleted when passed
the -faligned-new flag. Compilation of tcmalloc_unit fails in this case
because the aligned/sized new/delete functions were not defined and
the appropriate operators not overloaded.

Fixes #942.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -385,7 +385,7 @@ AM_CONDITIONAL(HAVE_SIZED_DEALLOCATION,
 AC_CACHE_CHECK([if C++ compiler supports std::align_val_t without options],
                [perftools_cv_have_align_val_t],
                [AC_LANG_PUSH(C++)
-                AC_COMPILE_IFELSE([AC_LANG_PROGRAM(
+                AC_LINK_IFELSE([AC_LANG_PROGRAM(
                     [[#include <new>]],
                     [[(::operator delete)((::operator new)(256, std::align_val_t(16)), std::align_val_t(16))]])],
                  perftools_cv_have_align_val_t=yes,
--- a/configure.ac
+++ b/configure.ac
@@ -397,7 +397,7 @@ AC_CACHE_CHECK([if C++ compiler supports -faligned-new],
                [AC_LANG_PUSH(C++)
                 OLD_CXXFLAGS="$CXXFLAGS"
                 CXXFLAGS="$CXXFLAGS -faligned-new"
-                AC_COMPILE_IFELSE([AC_LANG_PROGRAM(
+                AC_LINK_IFELSE([AC_LANG_PROGRAM(
                     [[#include <new>]],
                     [[(::operator delete)((::operator new)(256, std::align_val_t(16)), std::align_val_t(16))]])],
                  perftools_cv_have_f_aligned_new=yes,


alk:
Thanks a lot for the patch. Sadly I am not able to test on macos thus such issues occasionally happen :(

So the root of the problem seems to be that we test for -faligned-new when deciding to build and unittest this stuff, but we test for c++17 when we're providing functions in tcmalloc.h. So there might be some chance to hit this problem on other OSes then.

Let me CC @Lastique who authored original support.

And of course we still lacked those operators for OSX-specific operators overrides.

As for the patch, unfortunately relying on ENABLE_ALIGNED_NEW_DELETE in tcmalloc.h (which we install on users systems) isn't great idea.

Give me a bit to think about best way to address it. Or maybe I'll realize something that I am not seeing yet.

alk:
Yes, I was wrong above. That we test only for C++17 in tcmalloc.h is harmless.

The reason for test failing is that MacOS doesn't _have_ those aligned new/delete operators at all (!). Funnily, they ship compiler with -faligned-new but fail to provide necessary runtime support.

MacOS has some weird way to hook malloc. Perhaps because we're trying not to break their 'pools' support. And that is also one reason why malloc is so much slower on macos than other sensible oses (e.g. issue #886). For that reason we're hooking deep into their malloc and not provide regular operators new/delete. Because of that I think we shouldn't be trying to default aligned new/delete if OS doesn't support them at all yet.

So I am wondering if better fix is to change AC_COMPILE_IFELSE when testing for aligned new to AC_LINK_IFELSE. I.e. to detect that OSX doesn't actually support -faligned-new. Can you please try that ?

Lastique:
AFAIR, my patch checked at configure stage that (a) the compiler supports `-faligned-new` and (b) that it provides `std::align_val_t`. If any of those is true then gperftools should be compiled with aligned `new`/`delete` support. Is the problem that the compiler does not define `__cplusplus` to C++17 in this case? If so, we should add `-std=c++1z`.

PS: As I mentioned in my PR, I only tested my patch with gcc on Linux. I have no way to test it on OS X.

zuercher:
Apparently on this compiler `-std=c++1z` is still c++ 14:

```
$ c++ -std=c++1z -Isrc -dM -E src/common.cc | grep plusplus
#define __cplusplus 201406L
```

I'll try the AC_LINK_IFELSE idea next.

Lastique:
The test for `std::align_val_t` should probably be converted as well.

zuercher:
I'll convert that check as well, though it fails even without the link step because `std::align_val_t` only becomes available when `-faligned-new` is set.

alk:
applied. Thanks a lot.

Issue #919, opened 2017-08-19
Can't get java stack frame when trace off-heap memory


louxiu:
1. Problem: The java stack frame is missed(run program with `XX:+PreserveFramePointer`) when profile heap, but it can be got using `perf-java-flames`
![pprof_missing](https://user-images.githubusercontent.com/925154/29485338-04e70534-8503-11e7-9a41-7356c9f910f4.png)
The java part is missing in above pprof picture.[original pprof svg file](https://www.dropbox.com/s/mts7raidvtvk69q/204f2e60_leak.new.pprof-test.svg?dl=0), [original heap data file](https://www.dropbox.com/s/kuxun69w0ip1epp/cfdaccce_pprof.0001.new.heap?dl=0)
![perf_keeped](https://user-images.githubusercontent.com/925154/29485343-07d0ce74-8503-11e7-8b5d-f3295243e9fb.png)
The java part is kept in above perf picture.[original perf svg file](https://www.dropbox.com/s/uoxp35mr5b4huuz/4ca716ea_flamegraph-171024.svg?dl=0), [original perf data file](https://www.dropbox.com/s/yf247xa3my5l0qx/224d42c8_perf-171024.data?dl=0)
2. Reproduce: 
`java -XX:+PreserveFramePointer -jar pprof-test.jar`
[The simple maven project](https://www.dropbox.com/s/h0mukr3syb10tuv/pproftest.tar.gz?dl=0)
[The jar file for convient](https://www.dropbox.com/s/ln7ut35me3a2rla/pprof-test.jar?dl=0)

3. Versions: java version "1.8.0_111",  gperftools-2.6.1, libunwind-1.2.1 and lastest pprof (`go get github.com/google/pprof`)
Though I installed and `ldconfig` libunwind, but when `./configure` in gperftools I got `configure: WARNING: No frame pointers and no libunwind. Using experimental backtrace capturing via libgcc. Expect crashy cpu profiler.` So I `--enable-frame-pointer`, then it seems ok.
 4. Env:  `Linux  3.10.0-229.el7.x86_64 #1 SMP Fri Mar 6 11:36:42 UTC 2015 x86_64 x86_64 x86_64 GNU/Linux`

louxiu:
@aalexand maybe interested in this.

alk:
The issue is that gperftools format for profiles does not support JIT-ted code. It will be a bit of a major project to support, which is unlikely to happen any time soon.

ceeaspb:
[...]

PR #966, opened 2018-03-19
add option to add jit symbol map

louxiu:
The perf map is fetched using [create-java-perf-map.sh](https://github.com/jvm-profiling-tools/perf-map-agent/blob/master/bin/create-java-perf-map.sh)

Use command and attachment below to test.
`pprof --jit_map_file=perf-12114.map --svg /usr/bin/java heapprof.0001.heap > output.svg`
[heapprof.0001.heap](https://www.dropbox.com/s/0o80hq18kqxolnu/heapprof.0001.heap?dl=0)
[perf-12114.map](https://www.dropbox.com/s/yolnadc1hqlcfuj/perf-12114.map?dl=0)


alk:
As noted on: https://github.com/gperftools/gperftools/issues/919#issuecomment-385287986 this looks plausible but perf pprof is deprecated and you should make same change to golang pprof.

Could be standalone tool that eats protobuf-format of profile and symbolizes jit-ful functions.

PR #970, opened 2018-03-26
Support userland ASLR on 64 bits Intel architecture Linux

LittleCVR:
In this configuration, we are not constrained by the address space and we don't mind fragmentation. But to be on the safe side, we only ever fragment half of the address space.

Reference:
https://bugs.chromium.org/p/chromium/issues/detail?id=170133
https://bugs.chromium.org/p/chromium/issues/detail?id=173371

### DIFF
--- a/src/system-alloc.cc
+++ b/src/system-alloc.cc
@@ -107,6 +107,140 @@ bool CheckAddressBits(uintptr_t ptr) {
   return (ptr >> kAddressBits) == 0;
 }
 
+#if defined(OS_LINUX) && defined(__x86_64__)
+#define ASLR_IS_SUPPORTED
+#endif
+
+#if defined(ASLR_IS_SUPPORTED)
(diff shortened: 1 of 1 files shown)


alk:
Looking at this closer, I have couple concerns about code. But biggest question I have is whether this is really necessary on modern Linux kernels. I'll need a bit more time for this.

alk:
Indeed, what I am seeing on my machine is that mmap address is very well randomized. What google3 mmap allocator is recently doing (grab large area via mmap with PROT_NONE and then mprotect parts as we need), seems to be best option. Perhaps it is time to port that behavior.

So there seem to be no reason at all doing this kind of ASLR. Do you agree? I am thinking, if we agree, it should be dropped in Chromium as well.

alk:
Thanks for the patch, but given description above, this is not relevant anymore. ALSR doesn't need malloc to do anything. And if I recall even chrome isn't doing it anymore.

PR #973, opened 2018-03-26
Don't let tcmalloc read all kinds of env vars at startup

LittleCVR:
The env vars are still used in profiling=1 builds, by such illustrious
figures as the deep memory profiler bot and willchan. Reading from the
environment variables uses static initializers, so this change will
eliminate those static initializers in profiling=0 builds.

As discussed in the chromium-dev thread "Does anyone use tcmalloc's
debugging envvars?".

Reference:
https://bugs.chromium.org/p/chromium/issues/detail?id=94925

### DIFF
--- a/src/base/commandlineflags.h
+++ b/src/base/commandlineflags.h
@@ -157,6 +157,8 @@ namespace tcmalloc {
 // These macros (could be functions, but I don't want to bother with a .cc
 // file), make it easier to initialize flags from the environment.
 
+#if defined(ENABLE_PROFILING)
+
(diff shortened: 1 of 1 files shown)


LittleCVR:
Do you want to change the default (not allowing tcmalloc to read env vars) or do you want to keep such behavior?

alk:
Yes, default should be kept as it is now.

alk:
Hi. Do you still want this merged? Again, I understand that chromium folks have forbidden static initializers and for them this is good thing to do, but for general gperftools we need those environment variable to work.

PR #974, opened 2018-03-26
Add total delay metric for spinlocks

acmorrow:
The work here is derived from changes made to the vendored copy of gperftools in MongoDB. For background on these changes, please see https://jira.mongodb.org/browse/SERVER-30114.

In our testing, we did not observe a significant performance impact from this instrumentation, as it only affects the spin lock slow path, but it is definitely a potential concern for the general case. We also had some discussion around whether the platform introspection here is sufficient - we rely on `clock_gettime` and `CLOCK_MONOTONIC` for all platforms of interest to us, but we recognize that there may be other platforms where these are not available.

We think though that on balance this diagnostic information is very useful for understanding certain types of workload issues, and we are interested in finding a way to merge it if possible.

### DIFF
--- a/src/base/spinlock.cc
+++ b/src/base/spinlock.cc
@@ -49,6 +49,47 @@ const base::LinkerInitialized SpinLock::LINKER_INITIALIZED =
     base::LINKER_INITIALIZED;
 
 namespace {
+
+const int64_t kNanoPerSec = uint64_t(1000) * 1000 * 1000;
+
+#ifdef _WIN32
+int64_t frequency;
+
+void InitTimer() {
+    LARGE_INTEGER large_freq;
+    QueryPerformanceFrequency(&large_freq);
+    frequency = large_freq.QuadPart;
+}
+
+int64_t NowMonotonic() {
+  LARGE_INTEGER time_value;
+  QueryPerformanceCounter(&time_value);
+  return time_value.QuadPart;
+}
+
+int64_t TicksToNanos(int64_t timer_value) {
+  return timer_value * kNanoPerSec / frequency;
+}
+
+#else // _WIN32
+void InitTimer() {}
+
+int64_t NowMonotonic() {
+  struct timespec t;
+
+  if (clock_gettime(CLOCK_MONOTONIC, &t)) {
+    return 0;
+  }
+
+  return (static_cast<double>(t.tv_sec) * kNanoPerSec) + t.tv_nsec;
+}
+
+int64_t TicksToNanos(int64_t timer_value) {
+  return timer_value;
+}
+
+#endif // _WIN32
+
 struct SpinLock_InitHelper {
   SpinLock_InitHelper() {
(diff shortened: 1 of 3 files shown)


acmorrow:
@markbenvenuto and @bdlucas1 - Could you please relay to @alk the details behind the decision to use double here?

alk:
Overall this looks great. It is distinct from how Google-internal contention profiling works, but it seems fine as far as I can see. Lets just figure out that "double" thingy which looks odd.

acmorrow:
@alk - I've asked the original authors of the change to provide an update on the choice of double. 

Separately, per your comment about this being distinct from the internal google contention profiling, I've seen various hints dropped that gperftools is at this point fairly divergent from the internal state of the allocator. Could you provide some insight into what the plan is? Will they re-converge? Will gperftools continue to diverge? Will the current internal state ever be released as open source?

I'm becoming interested in this now because we are looking to upgrade MongoDB from gperftools 2.5 to whatever is latest, probably sometime this summer. As we have several modifications to our vendored copy, we are interested in any information that you can provide that will help us prepare for whatever may be coming.

alk:
Google-internal spinlock is already part of abseil. It actually has contention tracking, but contention profiler is amputated. Perhaps deemed not high-quality enough (it does have some issues).

As for Google-internal tcmalloc, yes stuff has diverged. And yes there are some concrete plans to open it too. But there are some dependencies like kernel support for rseq and also there are lots of unportable stuff. Like dependencies on internal-only glibc features. When this happens and which form is unknown. It is also unknown if end-result will be competitive with gperftools in portability.

So we should expect gperftools to remain for at least year or so.

alk:
I guess I can fix the double myself, return value is int64. Just please ack if you still want this. Also since we're not c++11 and later we can drop CAS-ing in stats adding code.

acmorrow:
We are definitely still interested in this. I've asked @markbenvenuto to follow up with background on the choice of double.

markbenvenuto:
One of the original internal code reviewers was worried about the possibility of integer overflow in the original code:
`return t.tv_sec * 1000000000ull + t.tv_nsec`

And while integer overflow could happen if the domain of the tv_sec value was (0, MAX_UINT64), given that tv_sec is just seconds ~584 years would need to pass before it would overflow. I placated the internal reviews by just adding the cast rather then argue.

Feel free to remove the `static_cast<double>`.

alk:
Hi all. This is still useful patch, but if you want it merged, please remove that double cast thing as agreed. Then I'll merge.

acmorrow:
@markbenvenuto - Can you rebase this and remove the double cast? We can always decide to re-apply it to our own copy when we next do an upgrade, or just decide not to worry about it.

alk:
You guys still want this ?

acmorrow:
Hi @alk - I'm no longer involved with MongoDB, so it isn't really my call. Perhaps @markbenvenuto will have more information about whether this is still relevant.

PR #979, opened 2018-04-07
Fix signature of sbrk.

cmuellner:
The manpage of sbrk says that its argument is of type intptr_t.
This patch fixes a compiler warning on gcc 7.3.0.

Signed-off-by: Christoph Müllner <christophm30@gmail.com>


alk:
Applied. Thanks. Indeed sbrk in glibc had intptr_t argument since at least 2000.

PR #980, opened 2018-04-07
linux: add aarch64_ilp32 support.

cmuellner:
This patch adds aarch64 (arm64) ILP32 support by using
the proper syscalls for sys_fcntl(), sys_fstat(), sys_lseek()
and sys_mmap().

Signed-off-by: Christoph Müllner <christophm30@gmail.com>

### DIFF
--- a/src/base/linux_syscall_support.h
+++ b/src/base/linux_syscall_support.h
@@ -2221,17 +2221,23 @@ struct kernel_stat {
         LSS_REG(0, arg1); LSS_BODY(type, name, "r"(__x0));                    \
       }
     #undef _syscall2
-    #define _syscall2(type, name, type1, arg1, type2, arg2)                   \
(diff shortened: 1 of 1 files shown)


cmuellner:
I think this patch needs some further information:

aarch64 (arm64) uses a LP64 ABI by default, but also offers a ILP32 ABI (similar to x86_64 and x32).
One of the benefits of aarch64_ilp32 is the potential performance benefit (e.g. pointer chasing workloads).
Using ILP32's performance benefits in combination with gperftools to further improve execution times
was the main motivation for this patch.

Details for ILP32 can be found here:
https://wiki.linaro.org/Platform/arm64-ilp32
https://wiki.debian.org/Arm64ilp32Port
https://wiki.leapproject.ca/index.php?title=Main_Page

alk:
Applied. Thanks.

PR #983, opened 2018-04-13
Update documentation for heap_checker.html

bdangit:
Make it clear that the static methods used require the "namespace", `HeapLeakChecker::`.


alk:
Merged. Thanks.

PR #985, opened 2018-04-15
Fix WIN32_OVERRIDE_ALLOCATORS for VS2017

HolyWu:
I have run all unittests (except preamble_patcher_test because I can't even compile it) under x86 & x64 and all passed. But I have uninstalled VS2015 long time ago so I can't confirm whether VS2015 works. But theoretically it should work as well since both use UCRT. The changes in `windows/config.h` are irrelevant to the fix. It's just that I want to synchronize the settings with main `config.h` for better comparison and maintenance.

### DIFF
--- a/src/libc_override.h
+++ b/src/libc_override.h
@@ -58,7 +58,7 @@
 #endif
 #include <gperftools/tcmalloc.h>
 
-#if __cplusplus >= 201103L
+#if __cplusplus >= 201103L || (defined(_MSC_VER) && _MSC_VER >= 1900)
 #define CPP_NOTHROW noexcept
 #define CPP_BADALLOC
 #else
--- a/src/tcmalloc.cc
+++ b/src/tcmalloc.cc
@@ -1691,6 +1691,10 @@ extern "C" PERFTOOLS_DLL_DECL int tc_set_new_mode(int flag) PERFTOOLS_NOTHROW {
   return old_mode;
 }
 
+extern "C" PERFTOOLS_DLL_DECL int tc_query_new_mode() PERFTOOLS_NOTHROW {
+  return tc_new_mode;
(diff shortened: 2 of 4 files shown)


src/windows/config.h
```diff
@@ -1,13 +1,11 @@
 /* A manual version of config.h fit for windows machines.
- *
- * Use of this source code is governed by a BSD-style license that can
- * be found in the LICENSE file.
- */
+
+   Use of this source code is governed by a BSD-style license that can
```
alk: Please, get rid of unrelated comment style changes.

alk:
Thanks for the patch. On high level it looks plausible. But there are lots of unrelated changes to windows/config.h. I would like to clean that file from unrelated changes.

Also, if possible, please update commit message with description of how you're fixing VS2017 compat.

Thanks again.

HolyWu:
Hi, I rebased the commit to remove the changes to `windows/config.h` and add some description to it. Regarding `windows/config.h`, will you clean it yourself or you want me to pull another request?

alk:
Much better. If you can do followup commit for config.h cleanup, that would be nice.

Can you also provide any documentation on how to do necessary libucrt "stripping"? I think I a lot of people would benefit greatly if you do.

HolyWu:
OK, I can do it in another commit.

I remove the objects from libucrt is just for "hide and seek". The tcmalloc users don't need to do the "stripping" anymore like before in order to use WIN32_OVERRIDE_ALLOCATORS, at least since VS2013 or even VS2010. VS2017 can build and run the unittests fine with the stock lib. It seems that Microsoft have defined heap-related functions as replaceable quite long time ago.

alk:
Merged. Thanks.

PR #986, opened 2018-04-17
Enable aligned new/delete declarations on Windows when applicable

HolyWu:


alk:
Thanks for the patch, but you also need to modify src/windows/tcmalloc.h.in (which autotools convert to tcmalloc.h you've modified). Perhaps we should consider moving TC_VERSION_XXX defines to own header so that (most of) tcmalloc.h is not autogenerated anymore.

HolyWu:
There is no file named `tcmalloc.h.in` at `src/windows/`, only `src/gperftools/` and `src/windows/gperftools/`. Can you confirm?

alk:
here: https://github.com/gperftools/gperftools/blob/master/src/windows/gperftools/tcmalloc.h.in

HolyWu:
Yes, that file is what I mean and is already included in the commit.

PR #987, opened 2018-04-19
Add tests for sized deallocation

HolyWu:


alk:
Merged. Thanks. I am not 100% happy with unrelated ifdef change, but decided not to block commit for that.

PR #992, opened 2018-04-29
Use off64_t instead of __off64_t

ffontaine:
__off64_t is a kernel internal symbol, which happens to be user-visible
with glibc, but not necessarily with other C libraries such as musl.

Switching from __off64_t to off64_t allows the code to build properly
with musl.

### DIFF
--- a/src/malloc_hook_mmap_linux.h
+++ b/src/malloc_hook_mmap_linux.h
@@ -60,7 +60,7 @@
 
 static inline void* do_mmap64(void *start, size_t length,
                               int prot, int flags,
-                              int fd, __off64_t offset) __THROW {
+                              int fd, off64_t offset) __THROW {
   return sys_mmap(start, length, prot, flags, fd, offset);
 }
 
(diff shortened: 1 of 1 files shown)


alk:
Thanks. I am wondering if we should prefer off_t instead. We'll simply need to ensure that right defines are given to make off_t 64-bit on 32 bit architectures.

On the other hand other forks of linux-syscall-support and abseil do use off64_t: https://github.com/abseil/abseil-cpp/blob/master/absl/base/internal/direct_mmap.h

alk:
I am currently trying to test on i386 GNU/Linux. Have you tested this on i386 musl?

ffontaine:
It works on musl with one more patch that I will send to you concerning src/malloc_hook_mmap_linux.h. In this file, __UCLIBC__ define is used instead of a check for __sbrk function in configure.ac, indeed __sbrk is specific to glibc, so it won't be available in uclibc or musl but musl does not define __MUSL__

alk:
So to clarify, you tested on i386 right? As sbrk, sure, testing it configure is fine. It is also fine to drop sbrk usage on musl.

alk:
Applied. Thanks.

PR #994, opened 2018-04-30
Make _recalloc adhere to MS's definition

HolyWu:
Although we still suffer from the fact that the value returned by tc_malloc_size could be larger than the actual requested size, it's the best what we can do I guess.

This essentially replaces a stale PR #714.

### DIFF
--- a/src/windows/override_functions.cc
+++ b/src/windows/override_functions.cc
@@ -66,10 +66,25 @@ void* _calloc_base(size_t n, size_t size) {
   return calloc(n, size);
 }
 
-void* _recalloc(void* p, size_t n, size_t size) {
-  void* result = realloc(p, n * size);
-  memset(result, 0, n * size);
-  return result;
+void* _recalloc(void* old_ptr, size_t n, size_t size) {
+  // Ensure that (n * size) does not overflow
+  if (!(n == 0 || (std::numeric_limits<size_t>::max)() / n >= size)) {
+    errno = ENOMEM;
+    return NULL;
+  }
+
+  const size_t old_size = tc_malloc_size(old_ptr);
+  const size_t new_size = n * size;
+
+  void* new_ptr = realloc(old_ptr, new_size);
+
+  // If the reallocation succeeded and the new block is larger, zero-fill the
+  // new bytes:
+  if (new_ptr != NULL && new_size > old_size) {
+    memset(static_cast<char*>(new_ptr) + old_size, 0, tc_nallocx(new_size, 0) - old_size);
+  }
+
+  return new_ptr;
 }
 
 void* _calloc_impl(size_t n, size_t size) {
--- a/src/tcmalloc.cc
+++ b/src/tcmalloc.cc
@@ -1391,7 +1391,7 @@ ATTRIBUTE_ALWAYS_INLINE inline void* do_calloc(size_t n, size_t elem_size) {
 
   void* result = do_malloc_or_cpp_alloc(size);
   if (result != NULL) {
-    memset(result, 0, size);
+    memset(result, 0, tc_nallocx(size, 0));
   }
   return result;
 }


alk:
Thanks. It it still has a problem with padding from allocated size and tc_malloc_size (the later is larger) not zeroed. But not sure if we can do anything about that.

alk:
Also should we handle case where oldptr is NULL ?

alk:
I think if we ensure that calloc zeroes all of the block (as seen by tc_malloc_size), then we should be good. Also consider handling oldptr == NULL case.

HolyWu:
`realloc` handles two special cases for us, which are also required by `_recalloc`: (old_ptr == NULL) and (new_size == 0). So no need to do redundant handling here.

alk:
So there is still a problem I think.

Consider something like this:

```
char* volatile deopt;

void* example() {
    char* t1 = new char[48];
    memset(t1, 0xac, 48);
    deopt = t1;
    delete [] deopt;
    char *t2 = static_cast<char*>(calloc(1, 47));
    // it is likely that t2 == t1
    // but last last byte is not zero-ed
    deopt = t2;
    // we'll see malloc_usable_size(t2) == 48,
    // will copy t2[47] as part of realloc, but not zero it.
    void *retval = _recalloc(deopt, 1, 128);
    assert(static_cast<char*>(retval)[48] == 0);
    return retval;
}
```

HolyWu:
Yes, that's why I mentioned "we still suffer from the fact" in the beginning. But I think it's still better than zeroing the entire range in all cases.

alk:
What do you mean by better? Correctness is clearly more important than performance or other factors, isn't it ?

alk:
Ah, you referred to original code being more broken. I'll take you patch and fix calloc. And then we'll be good 100% I think.

src/windows/override_functions.cc
```diff
@@ -66,10 +66,24 @@ void* _calloc_base(size_t n, size_t size) {
...
 
-void* _recalloc(void* p, size_t n, size_t size) {
-  void* result = realloc(p, n * size);
-  memset(result, 0, n * size);
-  return result;
+void* _recalloc(void* old_ptr, size_t n, size_t size) {
+  // Ensure that (n * size) does not overflow
+  if (!(n == 0 || (std::numeric_limits<size_t>::max)() / n >= size)) {
+    errno = ENOMEM;
+    return nullptr;
```
alk: We still support pre-c++11 compilers. Lets do NULL.

src/windows/override_functions.cc
```diff
@@ -66,10 +66,24 @@ void* _calloc_base(size_t n, size_t size) {
...
+  }
+
+  const size_t old_size = tc_malloc_size(old_ptr);
+  const size_t new_size = n * size;
+
+  void* new_ptr = realloc(old_ptr, new_size);
+
+  // If the reallocation succeeded and the new block is larger, zero-fill the
+  // new bytes:
+  if (new_ptr != nullptr && new_size > old_size)
```
alk: Rest of the code is adding {} braces around even single statement if bodies. Please do as as well.

alk:
Couple minor things inline. Otherwise good to go. Sadly, since this is override mode there is no easy way for me to even compile-test it.

HolyWu:
Rebased upon request. Out of curiosity, why you can't compile in override mode? Do you have problem with configuring MSVC projects?

alk:
Here is what I was about to do to calloc:

```
diff --git a/src/tcmalloc.cc b/src/tcmalloc.cc
index 48d4530..37c1440 100644
--- a/src/tcmalloc.cc
+++ b/src/tcmalloc.cc
@@ -1391,7 +1391,7 @@ ATTRIBUTE_ALWAYS_INLINE inline void* do_calloc(size_t n, size_t elem_size) {
 
   void* result = do_malloc_or_cpp_alloc(size);
   if (result != NULL) {
-    memset(result, 0, size);
+    memset(result, 0, tc_nallocx(size, 0));
   }
   return result;
 }
```

but then I noticed that recalloc also needs to do same to memset. Can you please make this change too? And feel free to include calloc change in your patch.

HolyWu:
Okay, rebased.

alk:
Applied. Thanks.

Issue #360, opened 2015-08-23
Option to disable writing tcmalloc larg allocation messages to stderr


alk:
Originally reported on Google Code with ID 357

```
tcmalloc will print messages such as the following to stderr:

tcmalloc: large alloc 1176436736 bytes == 0x3177a000 @  0x7f909f9fce6c 0x401ef9 0x40b931
0x44f1da 0x44be5f 0x4502c3 0x7f909ed9bc4d 0x401db9

Looking at the code, the output to stderr looks built in.  Is there a mechanism for
turning this off, as not to contaminate stderr unless there is actually an error?

```

Reported by `nilshomer` on 2011-07-29 19:43:15

ekg:
Would you be open to a PR doing this optionally, maybe with a compile flag?

alk:
Sure.

ekg:
I can verify that it is not possible to set this variable from within a binary. It must be set in the calling environment. This means that every distributed binary will need to be wrapped in a calling script if we would like to avoid this message.

I would vote for removing it. It's not like malloc displays warnings when allocating a moderate chunk of memory, and it's going to be more confusing to users than anything that tcmalloc does it.

That said, I now see how to add a compile option to do this...

alk:
I'd be open for compile-time option disabling it. I agree that it this automatic logging is perhaps a bit too opinionated, but I wouldn't be open (without stronger arguments at least) disabling it by default.

chenzheng1030:
Hi, is there any update for the compile-time option to disable the warning message?

alk:
Nothing changed recently. But I'll gladly accept patch to disable it by default.

alk:
Done. Thanks.

PR #1001, opened 2018-05-20
Disable large allocation report by default (#360)

jl2922:


jl2922:
#360

alk:
Thanks a lot. Applied.

PR #1004, opened 2018-06-04
Export GetStack* functions

### LINKED ISSUE #948: stacktrace.h and -lprofiler fails on OS X
I'm using functions from `<gperftools/stacktrace.h>` (and `"getpc.h"`) to regularly capture stack traces myself. I currently can't use `ProfilerStart()` and `ProfilerStop()`, because I need not only the frequency of each stack trace, but also the order of appearance (r-prof/gprofiler#18). Using `-lprofiler` seems to be sufficient on Ubuntu, but not on OS X. To reproduce, compile the following `test.cpp` file:

```
// clang test.cpp -lprofiler
// gcc test.cpp -lprofiler
#include <gperftools/stacktrace.h>

int main() {
  GetStackTraceWithContext(0, 0, 0, 0);
  return 0;
}
```

On OS X, I'm [seeing](https://travis-ci.org/r-prof/gprofiler/jobs/321992479#L474-L480):

```
Error: package or namespace load failed for ‘gprofiler’ in dyn.load(file, DLLpath = DLLpath, ...):
 unable to load shared object '/private/var/folders/bb/n7t3rs157850byt_jfdcq9k80000gn/T/Rtmp9E4hvH/Rinst182d22434e6c/gprofiler/libs/gprofiler.so':
  dlopen(/private/var/folders/bb/n7t3rs157850byt_jfdcq9k80000gn/T/Rtmp9E4hvH/Rinst182d22434e6c/gprofiler/libs/gprofiler.so, 6): Symbol not found: __Z24GetStackTraceWithContextPPviiPKv
  Referenced from: /private/var/folders/bb/n7t3rs157850byt_jfdcq9k80000gn/T/Rtmp9E4hvH/Rinst182d22434e6c/gprofiler/libs/gprofiler.so
  Expected in: flat namespace
 in /private/var/folders/bb/n7t3rs157850byt_jfdcq9k80000gn/T/Rtmp9E4hvH/Rinst182d22434e6c/gprofiler/libs/gprofiler.so
```

Compiling with `-ltcmalloc` resolves the issue. I wonder if this is expected. If not, happy to contribute.

On that note, perhaps `GetPC()` could be made available through `-lprofiler`, too? Or perhaps you'd support writing samples directly to file, without aggregation?

krlmlr:
Closes #948.

@alk: As discussed. I checked on OS X, it works for me. Thanks for considering.


krlmlr:
Does this need a comment why the `GetStack*` functions are exported?

krlmlr:
I signed the CLA.

alk:
Thanks for the patch. I am thinking we should improve this before we merge. Here is why.

So firstly, I am trying to be careful w.r.t. ABI compatibility issues. Which means APIs we introduce should be future-proof. Secondly, existing API has some concerns.

*) function names are a bit too generic, un-prefixed and un-namespaced. I.e. we're seriously risking colliding with some other library. Thankfully, abseil stuff is namespaced, so at least we won't conflict with them.

*) this API could be useful from C, but it is symbols are not extern "C".

*) I don't really understand 'Frames' versions. This adds internal complexity, but I'm not aware of any usages.

*) currently interaction of skip_count and ucontext is a bit of mess. x86 and ppc codes simply use ucontext as "helper" for unwinding. Never as starting point of unwinding. Which is kinda confusing. libunwind version actually tries to use ucontext for starting point of unwinding (and then ignores skip count).

What is exactly is your use case? I am wondering if instead of exposing existing questionable API, we simply introduce something super-simple. E.g. maybe gperftools_backtrace or something like that. This one due to simplicity would be fine to support going forward.

What you think?

krlmlr:
Thanks for getting back. #948 contains some context and a description of the use case, repeated below:

> I'm working on a project that aims at providing unified stack traces for R code that calls into native code. Each stack trace contains R functions when in R, and then seamlessly switches to native functions. R comes with a sampling profiler, I'm daisy-chaining it with gperftools's profiler and combine stack traces when profiling has finished. Initially I thought I could use `ProfilerStartWithOptions()` with a filter, but I discovered (much too late) that gperftools doesn't write the samples to disk in the order of appearance. I decided to manually collect stack traces and write to disk directly, because I wanted this to work with stable versions of gperftools. I borrowed relevant code from gperftools, see [feature branch](https://github.com/r-prof/gprofiler/tree/f-%236-combine).

The project is funded by the R Consortium where Google is a member, if that matters.

This patch only describes the "simplest" solution (with the drawbacks you mentioned). We could instead:

- Introduce an option that makes sure that samples are written as they occur, without compressing
- For each backtrace, write the sequence number of the relevant sample(s) in addition to the count
- Adapt libunwind to patch stack traces so that the R code is available directly to gperftools (I asked on their mailing list with little success: http://lists.nongnu.org/archive/html/libunwind-devel/2017-09/msg00002.html)

I'd greatly appreciate your help.

alk:
Thanks. So you need stacktrace capturing from signal handler. So we have two options:

a) somehow extend profiler to give features you want

b) export stacktrace capturing function useful for signal handlers.

Option (b) looks like needs both skip_count and ucontext. So at least lets do proper name, extern "C" and document skip_count and ucontext interaction. No need to refactor internal API, just build "for export" version on top of existing internal API.

alk:
Also out of curiosity, why you need exact order of samples?

krlmlr:
Are you suggesting to proceed on this pull request, and implement thin wrappers around the existing functions (instead of simply exporting them)?

Do we need to document in `cpuprofile.html` ? Would you support a PR that converts the documentation to Markdown?

alk:
I am proposing, instead of exporting all GetStackXXX function, we essentially export GetStackTraceWithContext, but under better name and with extern "C".

No need to (but nice to have to) refactor internal usages of this function.

alk:
Hi. Are you still interested getting this "in". Again, as noted above, lets just make sure ABI is sensible before we merge this.

krlmlr:
Thanks for the nudge. I think we still need this. Would you like me to propose the changes to the internals you described above?

alk:
Yes, please.

krlmlr:
Opened a new pull request to address this.

PR #1016, opened 2018-07-10
Support viewing profiles in callgrind with templates

bsilver8192:
The new regex looks for the -> separator instead of a space. This is
important when using --no_strip_temp, because (at least on my system)
the demangler likes putting spaces between template arguments and
in things like `foo<bar<baz> >`. I say "on my system" because that seems
like the kind of thing that might change once C++11 becomes widely
adopted.

### DIFF
--- a/src/pprof
+++ b/src/pprof
@@ -1310,7 +1310,7 @@ sub PrintCallgrind {
                            [$_, $1, $2] }
                      keys %$calls ) {
     my $count = int($calls->{$call});
-    $call =~ /([^:]+):(\d+):([^ ]+)( -> ([^:]+):(\d+):(.+))?/;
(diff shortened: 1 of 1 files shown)


bsilver8192:
Hmmmm, the failing test is of the code I touched, but it fails when I try it even before my change, before getting to the part I changed. Looks like master used to pass on Travis though. Could somebody confirm whether this is a problem with my change or a known problem with master?

bsilver8192:
Oh, never mind, I figured out how to run the tests. Apparently the dependencies in the Makefile are underspecified. However, I can't reproduce the failure. I ran the exact same `./autogen.sh && ./configure && make -j`getconf _NPROCESSORS_ONLN` && make check` as Travis and it passed. Any ideas on debugging?

bsilver8192:
Fair enough. I realized that after I finished debugging the perl one, but I'm OK with just keeping this around locally then.

PR #1027, opened 2018-07-31
Fix uninitialized memory use in sampler_test

brian-peloton:
Sampler's documentation states the following:
  C++03 requires that types stored in TLS be POD.  As a result, you must
  initialize these members to {0, 0, false} before using this class!

However, the test code wasn't doing that. MemorySanitizer and
UndefinedBehaviorSanitizer both failed because of it.

### DIFF
--- a/src/tests/sampler_test.cc
+++ b/src/tests/sampler_test.cc
@@ -361,7 +361,7 @@ double StandardDeviationsErrorInSample(
 }
 
 TEST(Sampler, LargeAndSmallAllocs_CombinedTest) {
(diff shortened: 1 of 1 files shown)


alk:
Applied. Thanks. Not sure why you're referring to POD btw.

brian-peloton:
I'm just quoting the comment in the existing code. C++03 doesn't have the concept of trivially constructible AFAIK, which is what I think is relevant, but then again the C++03 standard also doesn't have thread_local so I'm not too sure what the comment is referring to either.

PR #1028, opened 2018-08-02
Port GWP-ASan from internal TCMalloc.

morehouse:
GWP-ASan is a low-overhead sampling-based utility for finding
heap-use-after-frees and heap-buffer-overflows.

This implementation supports up to 64 allocations at a time with sizes
in the range [1, kPageSize]. The sampling rate is controlled by the
environment variable TCMALLOC_GUARDED_SAMPLE_PARAMETER, which defaults
to never doing a guarded allocation.

When a memory error (segfault) is detected, stack traces are printed for
point of allocation, point of deallocation (if applicable), and point of
error.  This implementation crashes on segfault.


alk:
Installing segfault handler all the time is a problem. We cannot afford installing signals in a library in general case. Doing this if user explicitly asked for it is fine, but not silently.

go/gwp-asan link is irrelevant externally.

We'll also need to ifdef new code to gcc (because of builtin_ctzll) and some kind of unix or just Linux for now due to usage of mmap/mprotect/segfault signal etc.

I'll deal with map_noreserve and spinlock thingy internally first.

alk:
Please also add guarded_page_allocator_test and guarded_page_allocator_test.sh to .gitignore

alk:
Test is not ifdefed. But I am wondering if we can get cleaner ifdefing if include file isn't used when not supported. Same could be done to makefile. Perhaps even even via configure flag, but I am thinking perhaps asking you to do autoconf is too much. So lets just do something that works and I'll clean up as necessary.

Also travis tests are timing out. I fixed c++11 so it should be something else now.

alk:
Much thanks for your work, again. Test is not passing on my box (fairly regular up-to-date debian sid x86-64 machine). http://paste.debian.net/1037862/

And when I manually run test number 392 (one that fails), I see that it crashes without printing gwp-asan stuff.

alk:
On Tue, Aug 14, 2018 at 9:21 AM, Matt Morehouse <notifications@github.com>
wrote:

> Hmm. The failing tests are the ones that test for guarded sampling.
>
> Any way I can reproduce to debug? Have you disabled sampling in your
> configuration?
>

This is just standard ./autogen.sh && ./configure && make check.

And I just tested on work glinux machine and I get same results. So you
should be able to reproduce on yours.

alk:
On Tue, Aug 14, 2018 at 12:28 PM, Aliaksey Kandratsenka <
alkondratenko@gmail.com> wrote:

>
>
> On Tue, Aug 14, 2018 at 9:21 AM, Matt Morehouse <notifications@github.com>
> wrote:
>
>> Hmm. The failing tests are the ones that test for guarded sampling.
>>
>> Any way I can reproduce to debug? Have you disabled sampling in your
>> configuration?
>>
>
> This is just standard ./autogen.sh && ./configure && make check.
>
> And I just tested on work glinux machine and I get same results. So you
> should be able to reproduce on yours.
>

Actually, that was wrong. It passes with make check. Fails when running .sh
manually, but only due to lack of sampling environment variable.

So pastebin above is likely invalid. I'll double check on my home box later
today.

morehouse:
Looks like a bug in GCC:  https://gcc.gnu.org/bugzilla/show_bug.cgi?id=83428

alk:
Could be bug, but has to be a lot more subtle than one you linked. That ticket looks invalid to me.

alk:
Here is godbolt: https://godbolt.org/g/EWH1nY

Again, I am not sure I am able to read legaleze of standard but constexpr constructor for SlotMetadata "fixes" it.

morehouse:
Thanks for looking into this.  After looking through the standard, I have a theory about what's going on.

According to 7.1.5/4, constexpr constructors must initialize all data members, and any constructors called from the constexpr constructor must also be constexpr.

Now SlotMetadata doesn't technically have a constexpr constructor, but there's no reason its contents can't be initialized statically.  Some compilers will detect this and essentially insert the constexpr specifier on the SlotMetadata constructor, but compilers are not required to do this.

Now compilers are not required to emit errors if a constexpr constructor is not actually constexpr, and can instead silently remove the constexpr specifier.  However, if we try to instantiate an object as constexpr and there's no valid constexpr constructor, then the compiler must emit a compile error.  I will add a test for this to prevent future regressions.

To fix the issue, we can either add constexpr constructors to SlotMetadata and StackTrace as you suggested, or we can use aggregate initialization for SlotMetadata.  Aggregate initialization will not technically call a constructor, so we don't need to worry about constructors being constexpr.

alk:
Matt, looks like we should be ready for another round? Can you please update this to very latest internal code and post? Alternatively, we wait for tagged address sampling, but I think there is value shipping current version.

morehouse:
Any idea what the timeline is for re-open-sourcing TCMalloc?

Also, mitchp@ is starting to work on a version of GWP-ASan that lives in llvm/compiler-rt.  Short-term plan is to integrate that version with Scudo.  Long-term, we'd like to converge implementations between Google3, Chrome, and compiler-rt and just use the compiler-rt one for everything.

Maybe we can hold off on this PR for now?

alk:
Matt, you probably don't care about this anymore?

morehouse:
No plans to work on this.  We'll get GWP-ASan for free when the new TCMalloc is open sourced.

alk:
And it already has. But gperftools will stay. Which means it won't get
gwp-asan yet. But inho no big deal. Thanks for response.

PR #1030, opened 2018-08-09
Fix page_heap_test flakiness

brian-peloton:
After 71fa9f87 (use 2-level page map for 48-bit addresses), there is a
potential for allocs of additional leaves in the middle of the test.
With ASLR enabled on amd64 (without TCMALLOC_SMALL_BUT_SLOW), this
results in the test failing sometimes (< 1% of the time). Instead, make
sure all those leaves are allocated before deciding on the baseline heap
usage in the test.

### DIFF
--- a/src/tests/page_heap_test.cc
+++ b/src/tests/page_heap_test.cc
@@ -6,9 +6,13 @@
 // be found in the LICENSE file.
 
 #include "config_for_unittests.h"
+
+#include <stdio.h>
+
+#include <memory>
+
 #include "page_heap.h"
 #include "system-alloc.h"
-#include <stdio.h>
 #include "base/logging.h"
 #include "common.h"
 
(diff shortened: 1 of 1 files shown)


src/tests/page_heap_test.cc
```diff
@@ -64,7 +64,35 @@ static void TestPageHeap_Stats() {
...
 }
 
+// Allocates all the last-level page tables we will need. Doing this before
+// calculating the base heap usage is necessary, because otherwise if any of
+// these are allocated during the main test it will throw the heap usage
+// calculations off and cause the test to fail.
+static void AllocateAllPageTables() {
+  // Make a separate PageHeap from the main test so the test can start without
+  // any pages in the lists.
+  tcmalloc::PageHeap* ph = new tcmalloc::PageHeap();
```
alk: unique_ptr ?

src/tests/page_heap_test.cc
```diff
@@ -64,7 +64,35 @@ static void TestPageHeap_Stats() {
...
+  tcmalloc::Span *spans[20];
```
alk: Thanks for the patch. Can you add some comments about why 20 or 10? You want to ensure certain amount of metadata, then why we need several iterations? How can we ensure that test remains correct if we (or when we) change metadata data structures?

src/tests/page_heap_test.cc
brian-peloton: Done

I updated the other uses in this file too, to keep them the same.

src/tests/page_heap_test.cc
brian-peloton: It's a constant used in the actual tests down below. I pulled it out as a named constant to keep it in sync for the future instead of duplicating it more.

I'm doing the allocations in separate chunks to ensure the same metadata gets allocated as in the tests below by following the exact same pattern. I don't think that's guaranteed (especially for future changes) if I used fewer, larger chunks or something else.

brian-peloton:
Thanks for the review!

alk:
Applied. Thanks.

PR #1031, opened 2018-08-10
Clean up MSVC projects

HolyWu:


alk:
Thanks for the patch, but I don't quite understand what it is doing. Can you please add some description?  Is that adding config support for overriding in addition to patching? Something else in addition ?

HolyWu:
The messages are already in the commit message. Do I need to describe it more?

alk:
Ah, I missed that. Thanks. Merged.

PR #1048, opened 2018-10-02
Comfortable define

Scayre:
#define endl "\n"

#define in(x) insert(x)

#define sz(x) (int)x.size()


alk:
Thanks for the patch, but something is very wrong with this pull request. How come it adds so many commits that are already merged?  Also can you please state the purpose of those defines? Those seem against the "spirit" of google c++ style guide.

alk:
Something looks very broken with this specific pull request. Not sure exactly. I apologize for not looking at it sooner. I'll close pull request. If you want to propose it again, please open new one.

Also note that if you're trying to propose just few #define-s added and used in some files, I'll have to refuse such patch. Google C++ style guide (even more liberal older version) are advising against abusing macros. And frankly most people including myself agree. No reason and no value.

PR #1056, opened 2018-10-06
check for __sbrk

ffontaine:
__sbrk is not defined in musl so add a check for it in configure.ac and
replace the check on ifndef __UCLIBC__ by a check on ifdef HAVE__SBRK
in src/malloc_hook_mmap_linux.h


alk:
Thanks you very much. Applied.

PR #1057, opened 2018-10-07
add option to disable benchmarks

ffontaine:
unwind_bench uses getcontext function however it is not available on
musl (POSIX.1-2004 obsoleted this function and in POSIX.1-2008 it was
removed) so add --without-benchmarks option.


ffontaine:
Indeed, this is not needed anymore so I'll close this PR.

PR #1083, opened 2019-02-04
Fix build on FreeBSD

pkubaj:
On FreeBSD the following syntax is used:
mr REGA,REGB


alk:
Thanks for the patch. I am a bit confused by this change though. This file is supposed to be used on apple OS, not free bsd. Can you please explain how it gets used on your OS ?

pkubaj:
It's included in https://github.com/gperftools/gperftools/blob/master/src/stacktrace.cc#L123

alk:
Applied. Thanks. Note, I have no means of testing this. So trusting your word here.

Issue #1079, opened 2019-01-29
Make more internal constants configurable


henrikedin:
In MongoDB we’ve changed a couple of hard-coded constants in `src/common.h`. I think it would make sense to expose more of them as configurable parameters when calling `./configure`

What do you think? I’m happy to send pull requests. We can discuss the specific constant in the respective PR.

One thing that we would like is more options for page size. I noticed #906 was opened but never merged, presumable because of the advanced shell script using log2. I can open a PR for a simpler solution.

alk:
Sure. Go ahead.

PR #1085, opened 2019-02-11
Allow configuring page size to 4K, 8K, 16K, 32K, 64K, 128K and 256K

henrikedin:
As talked about in #1079 I'm unsure how you edit config.h for Windows, but I manually mirrored the changes in config.h from ./configure
Set the lowest allowed page size to 4K, do you want to allow smaller than that?

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -113,8 +113,9 @@ To build libtcmalloc with large pages you need to use the
 
    ./configure <other flags> --with-tcmalloc-pagesize=32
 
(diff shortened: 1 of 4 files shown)


alk:
Thanks for the patch. Can you please fix tests when  page size is set to 4k?

henrikedin:
Should be OK now. Hope you agree with just increasing kClassSizesMax.

alk:
Applied. Thanks a lot.

PR #1087, opened 2019-02-15
Mark malloc extension allocation as ignored for the LeakSanitizer

romange:


romange:
Calms "-fsanitize=address" mode

alk:
Thanks for patch and apologies for slow reaction. This looks sensible. Can you please combine all the patches ?

romange:
Hi @alk. Github has squash and merge option. It's not the same?

alk:
I want nice and sensible commit message. Github whatever thingy is likely to make it crap.

Ah, another thing for commit specifically. windows doesn't have "weak" AFAIR. But lsan might still be useful there. We'll probably want to ifdef this somehow.

romange:
@alk  Unfortunately, I do not have windows build to verify this patch. I've applied the requested fixes.

romange:
Hi @alk, ptal

romange:
Hi @alk , is something wrong with the PR ?

alk:
I need some more time. Thing about weak attribute is it doesn't change anything for shared libraries. So then link order determines which function actually gets used. Stuff specified earlier on command line (or implicitly) usually gets precedence. So we'll need to check if libtcmalloc.so could actually override lsan's function. Which would be undesirable.

alk:
Also looks like weak is not really supported on OSX. Eh...

romange:
I guess this change brings more work than value :D. It was meant for
convenience, we can drop it.

PR #1099, opened 2019-04-08
Fix _aligned_realloc in WIN32_OVERRIDE_ALLOCATORS mode

HolyWu:


alk:
Thanks again. Same here. Can you please add more details on why you're doing this?

HolyWu:
No problem. _msize and _expand are called by _aligned_realloc internally. The added unit test would fail if both of them are not overridden. To avoid linker errors we override the *_base functions as well.

alk:
Thanks for the patch, but somehow your new test actually fails under x86 (i.e. 32 bit) and release+patch mode.

HolyWu:
I can reproduce it too. For unknown reason the call to `_aligned_realloc` causes an access violation under x86 when patching, but works fine under x64 as well as x86/x64 when overriding. Some mysteries here...

PR #1100, opened 2019-04-08
Fix linking error with Win10 SDK in WIN32_OVERRIDE_ALLOCATORS mode

HolyWu:


alk:
Thanks for the patch. Can you please elaborate a bit on why recalloc_base needs to be overridden as well? Something is calling it ?

HolyWu:
Yes. _recalloc_base is called by CRT internally. The linker will complain about one or more multiply defined symbols found in libucrt if we don't override it.

PR #1105, opened 2019-04-17
Fix a long time typo

prehistoric-penguin:
Thought this fix may not effect the behavior or performance, but it seems indeed a bug.

The previous change for this line happened 10 years ago, via commit 6c3eaabd

```diff
@@ -4998,28 +5087,29 @@ sub GetProcedureBoundaries {
   # --demangle and -f.
   my $demangle_flag = "";
   my $cppfilt_flag = "";
-  if (system("$nm --demangle $image >$dev_null 2>&1") == 0) {
+  my $to_devnull = ">$dev_null 2>&1";
+  if (system(ShellEscape($nm, "--demangle", "image") . $to_devnull) == 0) {
```

### DIFF
--- a/src/pprof
+++ b/src/pprof
@@ -5322,7 +5322,7 @@ sub GetProcedureBoundaries {
   my $demangle_flag = "";
   my $cppfilt_flag = "";
   my $to_devnull = ">$dev_null 2>&1";
-  if (system(ShellEscape($nm, "--demangle", "image") . $to_devnull) == 0) {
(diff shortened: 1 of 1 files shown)


alk:
Applied. Thanks. But note that we're not maintaining perl version of pprof. So I could have rejected it too.

prehistoric-penguin:
So we'd better to use this tool?  I suggest adding some document in the README file.
https://github.com/google/pprof

alk:
Correct. github.com/google/pprof is recommended tool.

PR #1127, opened 2019-07-12
Remove executable flag for c++ files

prehistoric-penguin:
Remove executable flag for c++ files, their color looks ugly in console and misleading.


alk:
Great find. Thanks for the patch. Applied.

PR #1137, opened 2019-08-20
Add mips64* support

lina-not-linus:


lina-not-linus:
Added mips64 compilation option to code compilation

alk:
Thanks. Applied.

PR #1143, opened 2019-09-04
prefetch multi tc_slots from pageheap

flex1988:


flex1988:
@googlebot I signed it!

flex1988:
@googlebot I fixed it.

alk:
Thanks for patch. Couple things.

*) can you please describe (in commit message) why this is useful thing to do

*) rest of code in this project and c++ file puts '{' braces to be on same line as if/while/else etc. We're using Google C++ style. It won't be right for us to deviate from this. Can you please fix this ?

alk:
Gentle ping on the message above.

flex1988:
> Gentle ping on the message above.

 just seen a todo in docs, and don't have a strong reason to do this, so i will close it

PR #1165, opened 2020-01-20
tcmalloc: ability to disable fallback allocator in memfs

JonKohler:
Whenever tcmalloc memfs allocator runs out of huge memory, it falls back to the
default system allocator, which will start allocating from normal anonymous
memory pool (small pages).  For use cases that exclusively want hugepage backed
memory, such as using tcmalloc in combination with Intel SPDK, this is
problematic.

This change adds an environment variable "TCMALLOC_MEMFS_DISABLE_FALLBACK"
which when set to true will ensure that tcmalloc doesn't fallback to using the
system allocator when the system runs out of hugepages. This is false by
default, such that all other tcmalloc consumers will not see any changes.

For those curious, you can see part of our upstream SPDK work here: 
https://review.gerrithub.io/c/spdk/spdk/+/475943

### DIFF
--- a/src/memfs_malloc.cc
+++ b/src/memfs_malloc.cc
@@ -85,6 +85,10 @@ DEFINE_bool(memfs_malloc_ignore_mmap_fail,
 DEFINE_bool(memfs_malloc_map_private,
             EnvToBool("TCMALLOC_MEMFS_MAP_PRIVATE", false),
 	    "Use MAP_PRIVATE with mmap");
+DEFINE_bool(memfs_malloc_disable_fallback,
+            EnvToBool("TCMALLOC_MEMFS_DISABLE_FALLBACK", false),
+            "If we run out of hugepage memory don't fallback to default "
+            "allocator.");
 
 // Hugetlbfs based allocator for tcmalloc
 class HugetlbSysAllocator: public SysAllocator {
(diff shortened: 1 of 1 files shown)


JonKohler:
@googlebot I signed it!

alk:
Thanks for the patch, but I don't understand what you're trying to achieve. Can you elaborate? (In commit message is best)

Fallback gets enabled when attempt to allocate from memfs allocator fails, so at this point we know that we just cannot keep using it. If not fallback than what ?

alk:
Ah, looks like you want to OOM the program. Note that we already have memfs_malloc_abort_on_fail which will abort when hugetlb memory runs out. Yes, not same as OOM, but in practice most programs will crash on oom anyways.

If indeed you have valid use case for this mode, sure, I'll accept this patch. But please make sure to document what you're trying to achieve.

JonKohler:
Thanks @alk - I've asked one of our devs to help double check me from his perspective, to make sure we articulate it properly here, I'll wait for his ping back.

Also, with the recent "other tcmalloc" release, should we be thinking about porting this over there? or leave here in gperftools. We dont (and probably wont for a while) use the other guy, but I figured I'd ask to check if there is an expectation about parallel PR's into both projects?

alk:
No such requirement. Also abseil tcmalloc amputated memfs anyways (it is internal only at least for now; looks like due to some api or "is it useful for people" concern). You might do them favor to file ticket that memfs is useful.

alokkataria:
> Ah, looks like you want to OOM the program. Note that we already have memfs_malloc_abort_on_fail which will abort when hugetlb memory runs out. Yes, not same as OOM, but in practice most programs will crash on oom anyways.
> 
> If indeed you have valid use case for this mode, sure, I'll accept this patch. But please make sure to document what you're trying to achieve.

Hi @alk  I am looking for the following behaviour -
1. When a program requests memory with memfs initialized, it should always get huge page backed memory, if huge pages are not available it should return NULL. Hence disabling the fallback allocator path is important for us.
2.  Also when we run out of huge pages we should let the program requesting the memory handle that failure rather than crashing right there. One of the use case is our caching layer which when sees that we are running out of memory can start evicting out stale objects from the cache to make way for new objects. 

memfs_malloc_abort_on_fail doesn't help with both the requirements hence I added this new flag.

Thanks !

alk:
Applied. Thanks.

But note that it is super-super tricky to actually reliably use this in practice. Best approach is perhaps indeed using new handler and tc_set_new_mode, so that malloc also calls it, before returning nullptr.

Otherwise, all kinds malloc calls in code you don't control but call into is likely to fail when it sees oom.

And doing new handler right and safe is also very nontrivial thing to maintain. So be careful there.

Also note that since this behavior is not covered by tests, I cannot promise it won't get broken. So please consider contributing unit test.

PR #1166, opened 2020-01-20
tcmalloc: fragmentation overhead instrumentation

JonKohler:
This patch adds visibility into the overhead due to fragmentation for each size
class in the tcmalloc central free list, which is helpful when debugging
fragmentation issues.


JonKohler:
@googlebot I signed it!

alk:
Thanks for patch. Looks great. Can you please combine to single commit and rebase on top of master. We keep our history clean and it is unlikely somebody would care about your "fat fingered" fix few years from now. Much better if you squash those two commits together.

JonKohler:
ok, for the sake of keeping this super clean, I made a whole new patch and PR, with a single commit here: https://github.com/gperftools/gperftools/pull/1173. I'll just shut this one down in favor of that

PR #1170, opened 2020-02-07
Fix the removed std::allocator::pointer member type removed in C++20

stac47:
This member type was deprecated in C++17 and removed in C++20.


alk:
Applied. Thanks.

PR #1173, opened 2020-02-18
tcmalloc: fragmentation overhead instrumentation

JonKohler:
This patch adds visibility into the overhead due to fragmentation for each size
class in the tcmalloc central free list, which is helpful when debugging
fragmentation issues.


alk:
Thanks a lot. Applied.

PR #1175, opened 2020-02-25
New ProfilerGetStackTrace()

### LINKED ISSUE #948: stacktrace.h and -lprofiler fails on OS X

krlmlr:
Properly exported function that wraps `GetStackTraceWithContext()`, as discussed in #1004.

Closes #948.

@alk: I can't update #1004, let's continue discussion here.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -1313,7 +1313,7 @@ libprofiler_la_SOURCES = src/profiler.cc \
                          $(CPU_PROFILER_INCLUDES)
 libprofiler_la_LIBADD = libstacktrace.la libmaybe_threads.la libfake_stacktrace_scope.la
 # We have to include ProfileData for profiledata_unittest
-CPU_PROFILER_SYMBOLS = '(ProfilerStart|ProfilerStartWithOptions|ProfilerStop|ProfilerFlush|ProfilerEnable|ProfilerDisable|ProfilingIsEnabledForAllThreads|ProfilerRegisterThread|ProfilerGetCurrentState|ProfilerState|ProfileData|ProfileHandler)'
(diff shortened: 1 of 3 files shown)


alk:
Nearly forgot about this. Applied. Thanks.

krlmlr:
Thanks! When can I expect the update to hit the Debian and Homebrew repos?

Issue #1076, opened 2019-01-09
GCC 9: error: Stack Pointer register clobbered by 'rsp' in 'asm'


marxin:
New error in GCC 9, where one can't clobber rsp:

```
[   68s] In file included from src/base/linuxthreads.cc:52:
[   68s] ./src/base/linux_syscall_support.h: In function 'int local_clone(int (*)(void*), void*, ...)':
[   68s] ./src/base/linux_syscall_support.h:1486:75: error: Stack Pointer register clobbered by 'rsp' in 'asm'
[   68s]  1486 |                              : "rsp", "memory", "r8", "r10", "r11", "rcx");
[   68s]       |                                                                           ^
```

alk:
Thanks for reporting. Fixed now.

PR #1192, opened 2020-05-01
Don't try to mark esp as clobbered in linux syscall support.

rmacnak-google:
Compare 9e5b1628737c67b4587f937164572774592978c4, issue #1076


alk:
Thanks a lot. Applied.

PR #1201, opened 2020-06-28
Fix memleak on linux with kernel pagesize 64k

rosinL:
when release a memory region less than pagesize 64k, it
cannot release successfully resulted in memleak.

Signed-off-by: luo rixin <luorixin@huawei.com>

### DIFF
--- a/src/page_heap.cc
+++ b/src/page_heap.cc
@@ -519,7 +519,8 @@ Length PageHeap::ReleaseAtLeastNPages(Length num_pages) {
   // span from each list.  Stop after releasing at least num_pages
   // or when there is nothing more to release.
   while (released_pages < num_pages && stats_.free_bytes > 0) {
-    for (int i = 0; i < kMaxPages+1 && released_pages < num_pages;
+    int i = 0;
+    for (; i < kMaxPages + 1 && released_pages < num_pages;
          i++, release_index_++) {
       Span *s;
       if (release_index_ > kMaxPages) release_index_ = 0;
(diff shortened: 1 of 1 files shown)


src/page_heap.cc
```diff
@@ -519,7 +519,8 @@ Length PageHeap::ReleaseAtLeastNPages(Length num_pages) {
   // span from each list.  Stop after releasing at least num_pages
   // or when there is nothing more to release.
   while (released_pages < num_pages && stats_.free_bytes > 0) {
-    for (int i = 0; i < kMaxPages+1 && released_pages < num_pages;
+    int i = 0;
+    for (i = 0; i < kMaxPages+1 && released_pages < num_pages;
```
DerVogel2020, line 523: The initialization i=0 is not needed here.

alk:
Thanks for the patch, but I don't quite get exactly what it is fixing. Is that dealing with arms (or other machine types) with 64k hardware pages?

Then how exactly your patch is addressing the issue?

If we're dealing with 64k page architecture maybe better fix is to ensure that tcmalloc page size is 64k or more ?

rosinL:
yes, it is dealing with 64k hardware pages. when  tcmalloc page size be configed with 64k, it  also cannot release the mem region which is 64k but start address and end address crossing two hardware pages.  I am not sure the tcmalloc page size could be configed with 128K or more.

alk:
>> when tcmalloc page size be configed with 64k, it also cannot release the mem region which is 64k but start address and end address crossing two hardware pages

I don't think this it should be possible. If that happens, then it is a bug. Better fix that bug then trying some workarounds like your patch appears to propose.

rosinL:
ok, i will try to reproduce the case and fix the bug.

rosinL:
I am using ceph with tcmalloc. I change the tcmalloc page size to 64k and cann't reprodeuce the case(may be I am wrong).
I found  the process RES is 3.6g, but the tcmalloc stats show 5.0g is used.  Does the trem "Bytes in use by application" means the size of virtual memory? Is there mem wastes when config the tcmalloc page size to  64k？ @alk 
RES
tcmalloc

alk:
There are 2 known possibilities why RSS can be less than "in-use-by-app" metric in malloc.

1) swap. I.e. usage is indeed higher, but some process ram is paged out

2) app allocated some memory but didn't touch it (yet). In practice this most likely to happen when some code pre-allocates large buffer (say 1 meg) but then ends up using only fraction of it. Then only used fraction will be actually mapped to real pages in hardware and thus RSS will be less then malloc thinks it should be.

Issue #1194, opened 2020-05-11
Make kMaxStackDepth configurable


kerrick-lyft:
I'm profiling a program which has a stack depth of more than 64 frames. When a stack of over 64 frames is seen, the profiler will omit the highest frames in the call stack in order to limit the overall call stack size to 64 frames. This behavior interacts poorly with pprof's flamegraph visualization (see https://github.com/google/pprof/issues/537), and I would like a way to generate a profile with full stacks.

In gperftools the max call stack size is a constant and is configured by [`kMaxStackDepth` here](https://github.com/gperftools/gperftools/blob/f47a52ce85c3d8d559aaae7b7a426c359fbca225/src/profiledata.h#L104).

It would be great if `kMaxStackDepth` were configurable at runtime, and/or if it were set by default to a value high enough that almost any non-recursive program won't hit the limit (e.g. 256).

## Repro instructions
(copied from [this issue](https://github.com/google/pprof/issues/537))

1. Save [this program](https://gist.github.com/kerrickstaley/03ce209596678d8f19938cb18ff64a4a) locally as `main.cc`. This program has a series of nested functions, `main` -> `func1` -> `func2` ... -> `func62`, and `func62` calls `square_sum` which does some work and also calls `square` which does more work.
2. Compile with `g++ -g -O0 main.cc`.
3. Run with `LD_PRELOAD="/usr/lib/libprofiler.so.0" CPUPROFILE=test.cpuprof ./a.out`.
4. Dump profile contents with `pprof -raw test.cpuprof`.

## Expected result:
Some frame stacks have at least 65 frames.

## Actual result:
All frame stacks have at most 64 frames.

alk:
Yes, having more frames would be nice. Seems hard to make it completely runtime tweakable, but I would accept patch to make it (statically) bigger.

kerrick-lyft:
I opened a PR: https://github.com/gperftools/gperftools/pull/1203

alk:
Applied. Thanks a lot.

PR #1203, opened 2020-07-08
Increase kMaxStackDepth to 256

kerrick-lyft:
Some programs may have a stack depth of over 64. I encountered this issue when debugging a C++ module that was being called from a Python program—the Python call stack occupied most of the 64 frames and I couldn't see what was happening in C++.

Increasing kMaxStackDepth to 256 is more than enough for the use case described above and Ought To Be Enough For Anybody™. This should have a negligible effect on memory usage if the program doesn't actually use more than 64 frames.

Fixes https://github.com/gperftools/gperftools/issues/1194 .

### DIFF
--- a/src/profiledata.h
+++ b/src/profiledata.h
@@ -101,7 +101,7 @@ class ProfileData {
     int      frequency_;                  // Sample frequency.
   };
 
-  static const int kMaxStackDepth = 64;  // Max stack depth stored in profile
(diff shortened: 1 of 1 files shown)


kerrick-lyft:
@alk could you review?

alk:
Applied. Thanks a lot. I hope you don't mind, I tweaked amount to 254 words. Reason is there are 2 words in each entry for count and depth. So then total sizeof(Entry) becomes nice round number.

PR #1214, opened 2020-09-12
Cmake build

okhowang:
based on #938
and update `CMakeLists.txt` with latest `configure.ac` `Makefile.am`
TODO:
+ [ ] dist package
+ [ ] rpm package
+ [ ] deb package
+ [ ] some ctest failed on my system. (I'm not familiar with gperftools's test)


okhowang:
@googlebot I consent

alk:
Thanks. Merged as is with minimal verification. I'll be adding a note that this is not yet officially maintained build configuration of gperftools.

As for various deb and rpm packaging things, I advise to ignore those. make dist is well probably also not necessary in cmake world.

I would appreciate if someone gives me command line or several so that I can integrate it with CI system. That would ideally build and test several build configuration.

PR #1223, opened 2020-10-10
docs: fix simple typo, defininitions -> definitions

timgates42:
There is a small typo in src/base/linux_syscall_support.h.

Should read `definitions` rather than `defininitions`.


alk:
applied. Thanks.

PR #1231, opened 2020-11-17
Fix implicit-int-float-conversion warning.

PatriosTheGreat:
Clang 10 and above warm implicit cast from int to float, or double in this case.


PatriosTheGreat:
@googlebot I signed it!

alk:
Applied. Thanks!

PR #1233, opened 2020-11-27
Add OS X arm64 program counter

granthenke:
When building gperftools on OSX Big Sur (11.0.1) on the M1 processor (arm64) I saw the below message and the profiler was not built. 

> Could not find the PC.  Will not try to compile libprofiler...

To fix this I added the program counter that I could find on my machine.


granthenke:
@googlebot I signed it!

alk:
applied. Thanks.

PR #1239, opened 2020-12-24
Add tc_malloc_get_stats

bibmaster:
This function collects the same information as tc_malloc_stats for external logging / metrics.

### DIFF
--- a/src/gperftools/tcmalloc.h.in
+++ b/src/gperftools/tcmalloc.h.in
@@ -76,6 +76,20 @@
 # endif
 #endif
 
+struct tc_malloc_stats_t {
+  size_t thread_bytes;
+  size_t central_bytes;
+  size_t transfer_bytes;
+  size_t metadata_bytes;
+  size_t pageheap_system_bytes;
+  size_t pageheap_free_bytes;
+  size_t pageheap_unmapped_bytes;
+  size_t pageheap_committed_bytes;
+  size_t spans_in_use;
+  size_t heaps_in_use;
+  size_t page_size;
+};
+
 #ifdef __cplusplus
 extern "C" {
 #endif
@@ -103,6 +117,7 @@ extern "C" {
   PERFTOOLS_DLL_DECL void* tc_pvalloc(size_t __size) PERFTOOLS_NOTHROW;
 
   PERFTOOLS_DLL_DECL void tc_malloc_stats(void) PERFTOOLS_NOTHROW;
+  PERFTOOLS_DLL_DECL void tc_malloc_get_stats(tc_malloc_stats_t* stats) PERFTOOLS_NOTHROW;
   PERFTOOLS_DLL_DECL int tc_mallopt(int cmd, int value) PERFTOOLS_NOTHROW;
 #if @ac_cv_have_struct_mallinfo@
   PERFTOOLS_DLL_DECL struct mallinfo tc_mallinfo(void) PERFTOOLS_NOTHROW;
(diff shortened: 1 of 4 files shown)


bibmaster:
@googlebot I signed it!

alk:
Thanks for the patch and apologies for somehow missing this earlier. Question, why existing malloc extensions api is not good enough for you? Why adding new function?

Adding functions is easy, but I am trying to make sure tcmalloc's ABI is small and maintainable. Yes, just adding your function is easy but I want to make sure it is well understood why it needs to be added. And also in this case we need to be careful w.r.t. exposing some internal implementation details and what we'll do when those change some time in future.

bibmaster:
To obtain the same information in existing API I have to call GetNumericProperty multiple times. Which involves a lot of strcmp, ExtractStats, pageheap_lock. With reference to API stability, these are generic TCMalloc metrics. I apply this patch in my builds without any modifications for at least 5 years.

bibmaster:
My use-case: tcmalloc with global alloc registry. Mostly all allocations are reflected in metrics cache.
Services check memory consumption periodically and log allocation map on virtual size growth.
```
--------------------------------------------------------------------------------
 spans in use:41249, thread heaps in use:24, page size:8192
--------------------------------------------------------------------------------
 21 007 929 312 : 19.57 Gb :   in use by application
              0 :      0 b : + in use by page heap freelist
      4 178 304 :  3.98 Mb : + in use by central cache freelist
      8 403 968 :  8.01 Mb : + in use by transfer cache freelist
      3 551 904 :  3.39 Mb : + in use by thread cache freelists
     26 423 488 :  25.2 Mb : + in use by malloc metadata
 21 050 486 976 :  19.6 Gb : = actual memory used (physical + swap)
      1 228 800 :  1.17 Mb : + bytes released to OS (aka unmapped)
 21 051 715 776 : 19.61 Gb : = virtual address space used
--------------------------------------------------------------------------------
         allocated         :  count   :   peak   : type
--------------------------------------------------------------------------------
...
      2 633 416 :  2.51 Mb :    68727 :  2.51 Mb : string
...
      7 153 820 :  6.82 Mb :    70867 :          : other
--------------------------------------------------------------------------------
 20 931 181 650 : 19.49 Gb :          :          : total registered allocations
--------------------------------------------------------------------------------
```

alk:
Apologies for delay. How about api like this instead?

```

class MallocExtension {
 public:
  struct PropsBatch { ... }; // opaque and expected to be held on stack briefly
};

// Any property supported by GetNumericProperty will be supported. All pointers must be valid until batch realize call.
void tcmalloc_batch_get_property(const char* property, size_t* value, bool* available, MallocExtension::PropsBatch* batch);
void tcmallo_batch_realize(PropsBatch* batch);

// then used like this

struct my_tcmalloc_props {
  size_t thread_bytes;
  size_t central_bytes;
...
};


my_tcmalloc_props p;
memset(&p, 0, sizeof(p));

MallocExtension::PropsBatch pb;
tcmalloc_batch_get_property("tcmalloc.thread_bytes", &p.thread_bytes, nullptr, &pb);
tcmalloc_batch_get_property("tcmalloc.central_bytes", &p.central_bytes, nullptr, &pb);
...
tcmalloc_batch_realize(&pb);
// and now use p.thread_bytes etc

```

Yes there will still be some strcmp-ares but we're talking merely half-dozen and they'll be outside of locks. And this gets generality of API and implementation has good chance of only fetching stats that were requested.

Issue #1224, opened 2020-10-11
allow custom stacktrace provider


romange:
Hi @alk ,

gperftool already provides customization of stacktrace providers via setenv. Unfortunately, libunwind code immediately crashes on aarch64 (see https://github.com/libunwind/libunwind/issues/68). Even more, gcc based stacktrace provider is using libunwind if it's linked into the binary. Therefore,  gperf profiler works on aarch64 only if it's linked without libunwind dependency.

There is [abseil based](https://github.com/abseil/abseil-cpp/blob/master/absl/debugging/stacktrace.h) stacktrace provider that has a similar signature as the one in gperftools. If we could programmatically set stacktrace provider we could use the one from abseil and not being dependent on a linker configuration of the binary. 

Do you mind if I send a PR that will allow setting stacktrace provider programmatically?

alk:
Yes, that would be useful.

romange:
the crash has been fixed on aarch64, closing.

PR #1240, opened 2021-01-18
Preliminary fix for #1224

romange:
Hi @alk , this is a partial  fix for https://github.com/gperftools/gperftools/issues/1224

Please tell me if I can proceed with the implementation or you have questions/reservations.

Thanks

### DIFF
--- a/src/gperftools/stacktrace.h
+++ b/src/gperftools/stacktrace.h
@@ -1,11 +1,11 @@
 // -*- Mode: C++; c-basic-offset: 2; indent-tabs-mode: nil -*-
 // Copyright (c) 2005, Google Inc.
 // All rights reserved.
-// 
+//
 // Redistribution and use in source and binary forms, with or without
 // modification, are permitted provided that the following conditions are
 // met:
-// 
+//
 //     * Redistributions of source code must retain the above copyright
 // notice, this list of conditions and the following disclaimer.
 //     * Redistributions in binary form must reproduce the above
@@ -15,7 +15,7 @@
 //     * Neither the name of Google Inc. nor the names of its
 // contributors may be used to endorse or promote products derived from
 // this software without specific prior written permission.
-// 
+//
 // THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
 // "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
 // LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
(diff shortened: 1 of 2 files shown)


alk:
Hi. This is in principle good enough. Just lets make sure new name is properly "namespaced". Either use actual C++ namespace or do something like tcmalloc_set_custom_whatever.

At higher level, I wonder if perhaps better path to solve your specific problem is taking this working backtracing implementation from abseil and applying it to gperftools. Then perhaps we won't need this customization feature.

alk:
I've just submitted better backtracing for 64-bit arms, so maybe this won't be needed? Please test and let me know.

romange:
@alk  sure will check and get back to you. How should I configure it exactly - with libunwind or without it ?

alk:
When libunwind is available, the code explicitly checks if libunwind is returning bogus stacktraces and then defaults to frame pointers.

So it is up to you, but I'd try to have libunwind in case it gets unbroken. Since frame pointers unwinding has notable issue unwinding from functions without frame pointers (typically various asm codes like memcpy). In general if you only need backtraces for stuff like growthz-s, malloc sampling or heap checker (which is borked on arms anyways), then frame pointers should be good enough I think. cpu profiling I'd try to prefer libunwind.

Also note that I found that at least on debian there is another version of libunwind. Android's fork. That seems to work fine on arms (!). It is just a matter of installing it and passing right flags to configure to use it.

romange:
Could it be that it somehow messed with libunwind linker dependencies?

I get 
```
/usr/bin/ld: third_party/libs/gperf/lib/libprofiler.a(stacktrace.o): in function `GetStackFramesWithContext_libunwind(void**, int*, int, int, void const*)':
stacktrace.cc:(.text+0x1ac): undefined reference to `_ULaarch64_init_local'
/usr/bin/ld: stacktrace.cc:(.text+0x1b8): undefined reference to `_ULaarch64_step'
/usr/bin/ld: stacktrace.cc:(.text+0x1d4): undefined reference to `_ULaarch64_get_reg'
/usr/bin/ld: stacktrace.cc:(.text+0x22c): undefined reference to `_ULaarch64_step'
/usr/bin/ld: stacktrace.cc:(.text+0x248): undefined reference to `_ULaarch64_get_reg'
/usr/bin/ld: stacktrace.cc:(.text+0x27c): undefined reference to `_ULaarch64_get_reg'
```

I configure with `--enable-libunwind --disable-heap-checker --disable-debugalloc --disable-heap-profiler --enable-static=yes`

alk:
This means libunwind is somehow not being linked correctly. if you give me config.log I may be able to help.

romange:
@alk  thanks, attaching both config.log and libprofiler.pc 

[libprofiler.txt](https://github.com/gperftools/gperftools/files/6043167/libprofiler.txt)
[config.txt](https://github.com/gperftools/gperftools/files/6043168/config.txt)

alk:
Is it possible that your libunwind is static library?

That seems like only possible explanation I have. I.e. configure checks for backtrace symbol is available when -lunwind is passed. But if -lunwind finds .a and backtrace is also available in libc (as it is with glibc), it will pass the configure check and assume that -lunwind works.

You can fix this manually by passing -lunwind_aarch64 or whatever it is as LIBS environment variable in configure command.

romange:
Confirming that `generic_fp` works on aarch64. libunwind will continue crashing due to https://github.com/libunwind/libunwind/issues/68 until the fix will propagate to Ubuntu.

Thanks Aliaksey!

PR #1249, opened 2021-02-09
atomicops: Remove Acquire_Store / Release_Load

vsrinivas:
gperftools' internal atomicops library included atomic
Release Load and Acquire Store operations; those operations
were unused and expressed ordering constraints that aren't
expressible in the C++ standard memory model.

Remove them, to make a transition to C++11 atomics easier
and to avoid confusing use of them.


alk:
Thank you! Applied.

PR #1250, opened 2021-02-12
profile-handler: prefer documented sigev_notify_thread_id in sigevent

sgn:
sigevent(7) is documented to have sigev_notify_thread_id as its member.
In glibc system, it's a macro expanded to the legacy _sigev_un._tid,
_sigev_un._tid is obviously an internal implementation detail as
signaled by its underscore prefix.

On Linux that use musl libc, sigev_notify_thread_id is also a macro, but
it's expanded to __sev_fields.sigev_notify_thread_id

Let's use the documented member if exists with a fallback to the
undocumented member instead.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -212,6 +212,7 @@ check_include_file("sys/cdefs.h" HAVE_SYS_CDEFS_H) # Where glibc defines __THROW
 
 check_include_file("unistd.h" HAVE_UNISTD_H)
 check_include_file("inttypes.h" HAVE_INTTYPES_H)
+check_include_file("linux/signal.h" HAVE_LINUX_SIGNAL_H)
 # We also need <ucontext.h>/<sys/ucontext.h>, but we get those from
 # AC_PC_FROM_UCONTEXT, below.
 
--- a/configure.ac
+++ b/configure.ac
@@ -203,7 +203,7 @@ AC_CHECK_FUNCS(sbrk)            # for tcmalloc to get memory
 AC_CHECK_FUNCS(__sbrk)          # for tcmalloc to get memory
 AC_CHECK_FUNCS(geteuid)         # for turning off services when run as root
 AC_CHECK_FUNCS(fork)            # for the pthread_atfork setup
-AC_CHECK_HEADERS(features.h)    # for vdso_support.h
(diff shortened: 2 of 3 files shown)


sgn:
@googlebot I signed it!

alk:
Thanks for patch. glibc actually has sigev_notify_thread_id macro, but in linux/signal.h include. Notably musl doesn't seem to have this header. So perhaps we detect presence of linux/signal.h include it and use sigev_notify_thread_id for glibc too.

alk:
Also BSD does support this macro too, using your existing include. So very nice. https://www.freebsd.org/cgi/man.cgi?query=sigevent&sektion=3&apropos=0

sgn:
On 2021-02-12 22:29:20-0800, "Aliaksey Kandratsenka (aka Aliaksei Kandratsenka)" <notifications@github.com> wrote:
> Thanks for patch. glibc actually has sigev_notify_thread_id macro,
> but in linux/signal.h include. Notably musl doesn't seem to have
> this header. So perhaps we detect presence of linux/signal.h include
> it and use sigev_notify_thread_id for glibc too. 

Sorry for being ignorance.
But, what is your preference in this case:
- rewrite this patch entirely: detect if linux/signal.h exists and
  include linux/signal.h and use sigev_notify_thread_id
  unconditionally
- Write another patch on top of this patch?

alk:
Hm, turns out linux/signal.h is basically unusable. It defines things that are conflicting with libc types (seeing this in both glibc and musl too). I've updated kernel bug at https://bugzilla.kernel.org/show_bug.cgi?id=200081

I think I'll take your change with minimal fix on top to unbreak it for glibc systems (since configure rightfully detects that linux/signal .h is broken and we end up with compile error trying to use that macro that is not defined anywhere). Your first change was most right one given that kernel headers bug.

alk:
Filed glibc bug as well: https://sourceware.org/bugzilla/show_bug.cgi?id=27417

alk:
Merged. Thanks a lot for the patch! Indeed musl compilation was borked without it. As noted above, I've amended your change. Making clear notice about this in commit message. Hope this is okay (I left authorship on you, but hopefully blames on bug(s) in this change I accepted on me).

sgn:
Thanks.

PR #1251, opened 2021-02-18
tcmalloc: Switch thread-safety annotations to support clang

vsrinivas:
tcmalloc contains its own copy of thread_annotations.h, wrapper
macros for static thread-safety analysis expressions. These thread-
safety expressions allow asserting (at compile time) that certain
locks are held or excluded or certain data is protected by specific
locks; they are checked at compile-time by recent versions of clang
or a gcc branch (https://gcc.gnu.org/wiki/ThreadSafetyAnnotation).

Convert the #if-guard and macro names from the no-longer-supported
gcc branch's defines & macros to the versions supported by recent
versions of clang.


alk:
Perfect. Thanks. Merged.

PR #1263, opened 2021-04-01
Advanced path from env generation

artpol84:
Currently, only the children output files will get the PID.
However if using this tool to analyze MPI applications, it is
possible that multiple processes not related with parent-child
relations are running on the same host.

The other option would be to do a printf-like thing
`HEAPPROFILE=/tmp/profile-%p`
where `%p` portion would be replaced with the PID.
However, having the PID in the file name is always a good idea IMO.

Signed-off-by: Artem Y. Polyakov <artpol84@gmail.com>

### DIFF
--- a/src/base/sysinfo.cc
+++ b/src/base/sysinfo.cc
@@ -206,6 +206,71 @@ extern "C" {
   }
 }
 
+// A set of primitives to trick with the environemnt variables
+// to mark child processes.
+// As gperftools is commonly loaded dynamically using LD_PRELOAD,
+// users are leveraging environment variables to communicate the parameters
+// to the corresponding gperftools library.
+// Specifically, a set of environment variables are used to specify the file
+// paths to be used during the operation (i.e. output profile filename).
+// When the application is using fork() to spawn more processes, it is important
+// to avoid conflicts where both the parent and a child area using the same
+// filename.
+// In order to do so, the parent reads the user-defined environment and toggles
+// the MSB of the first byte in it's value to indicate that all other processes
+// using this variable are children (and thus) must append a unique identifier
+// (PID) as a suffix to the corresponsing filename.
+#define EnvChildSignSet(enval) { enval[0] |= 128; }
+#define EnvChildSignDrop(enval) { enval[0] &= ~128; }
+#define EnvChildSignIsOn(enval) (enval[0] & 128)
+
+// HPC environment auto-detection
+// For HPC applications (MPI, OpenSHMEM, etc), it is typical for multiple
+// processes not engaged in parent-child relations to be executed on the
+// same host.
+// In order to enable gperftools to analyze them, these processes need to be
+// assigned individual file paths for the files being used.
+// The function below is trying to discover well-known HPC environments and
+// take advantage of that environment to generate meaningful profile filenames
+static bool QueryHPCEnvironment(char *hpc_suffix) {
+  char *enval;
+  int procid, jobid;
+
+  // Check for the PMIx environment
+  procid = EnvToInt(TC_ENV_PMIX_RANK, -1);
+  if (0 <= procid) {
+    // PMIx exposes the rank that is convenient for process identification
+    snprintf(hpc_suffix, PATH_MAX, TC_ENV_PMIX_SUFFIX "%d", procid);
+    // Use PMIx rank to differentiate
+    return false;
+  }
+
+  // Check for the Slurm environment
+  jobid = EnvToInt(TC_ENV_SLURM_JOBID, -1);
+  if (0 <= jobid) {
+    /* Slurm environment detected */
+    procid = EnvToInt(TC_ENV_SLURM_PROCID, -1);
+    if (0 <= procid) {
+      snprintf(hpc_suffix, PATH_MAX, TC_ENV_SLURM_SUFFIX "%d", procid);
+      // Use Slurm process ID to differentiate
+      return false;
+    }
+    // Need to add PID to avoid conflicts
+    return true;
+  }
+
+  /* Check for Open MPI environment */
+  enval = getenv("OMPI_HOME");
+  if (enval) {
+    return true;
+  }
+
+  // TODO: Detect other environments - MPICH
+
+  // No HPC environment was detected
+  return false;
+}
+
 // This takes as an argument an environment-variable name (like
 // CPUPROFILE) whose value is supposed to be a file-path, and sets
 // path to that path, and returns true.  If the env var doesn't exist,
@@ -230,15 +295,37 @@ extern "C" {
 // TODO(csilvers): set an envvar instead when we can do it reliably.
 bool GetUniquePathFromEnv(const char* env_name, char* path) {
   char* envval = getenv(env_name);
+  char hpcSuffix[PATH_MAX] = "";
+  char forceVarName[TC_ENV_MAX_NAME];
+  bool wantPid = false;
+
   if (envval == NULL || *envval == '\0')
     return false;
-  if (envval[0] & 128) {                  // high bit is set
-    snprintf(path, PATH_MAX, "%c%s_%u",   // add pid and clear high bit
-             envval[0] & 127, envval+1, (unsigned int)(getpid()));
-  } else {
-    snprintf(path, PATH_MAX, "%s", envval);
-    envval[0] |= 128;                     // set high bit for kids to see
+
+  // Get information about the child bit and drop it
+  const bool isChild = EnvChildSignIsOn(envval);
+  EnvChildSignDrop(envval);
+
+  wantPid = QueryHPCEnvironment(hpcSuffix);
+
+  // Generate the "forcing" environment variable name in a form of
+  // <ORIG_ENVAR>_USE_PID that requests PID to be used in the file names
+  snprintf(forceVarName, TC_ENV_MAX_NAME, "%s_USE_PID", env_name);
+
+  if (isChild || EnvToBool(forceVarName, false)) {
+    // force PID usage
+    wantPid = true;
   }
+
+  if (wantPid) {
+    snprintf(path, PATH_MAX, "%s%s" TC_ENV_PID_SUFFIX "%d",
+             envval, hpcSuffix, getpid());
+  } else {
+    snprintf(path, PATH_MAX, "%s%s", envval, hpcSuffix);
+  }
+
+  // Set the child bit for the fork'd processes
+  EnvChildSignSet(envval);
   return true;
 }
 
--- a/src/tests/unique_path_unittest.cc
+++ b/src/tests/unique_path_unittest.cc
@@ -1,0 +1,288 @@
+/* -*- Mode: C++; c-basic-offset: 2; indent-tabs-mode: nil -*-
+ * Copyright (c) 2023, gperftools Contributors
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are
+ * met:
+ *
+ *     * Redistributions of source code must retain the above copyright
+ * notice, this list of conditions and the following disclaimer.
+ *     * Redistributions in binary form must reproduce the above
+ * copyright notice, this list of conditions and the following disclaimer
+ * in the documentation and/or other materials provided with the
+ * distribution.
+ *     * Neither the name of Google Inc. nor the names of its
+ * contributors may be used to endorse or promote products derived from
+ * this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+// ---
+// Author: Artem Y. Polyakov
+
+#include <stdlib.h>   // for environment primitives
+#include <unistd.h>   // for getpid()
+#include <limits.h>   // for PATH_MAX
+#include <string>
+#include <utility>
+#include "addressmap-inl.h"
+#include "base/logging.h"
+#include "base/sysinfo.h"
+#include "base/commandlineflags.h"
+
+#define TEST_VAR   "GPROF_TEST_PATH"
+#define TEST_FORCE TEST_VAR "_USE_PID"
+#define TEST_VAL   "/var/log/some_file_name"
+#define HPC_RANK "5"
+
+#define PID_SUFFIX TC_ENV_PID_SUFFIX
+#define PMIX_RANK_ENV TC_ENV_PMIX_RANK
+#define PMIX_SUFFIX TC_ENV_PMIX_SUFFIX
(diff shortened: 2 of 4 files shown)


artpol84:
@googlebot I signed it!

alk:
Apologies for delay. I guess I am finally okay with just doing this. Wasn't sure if changing behavior is fine or not. But I have a comment about change itself.

So now that we always do pid thing, doesn't look like we need "or bit 7" trick anymore. Can you please clean up this part? (Or did I misundertand something)

artpol84:
Yes, I don't think we need the extra bit trick anymore. Let me have a look. I assume we need to avoid parents setting it. So it'll be in a different place.

artpol84:
> Can you please clean up this part? (Or did I misundertand something)

Done cleaning up and updating the PR.

alk:
Eh. I am very very sorry. Looking at this once more I cannot agree with "always append pid" approach. For most people and me too, main workload with cpu profiler is something like:

```
$ CPUPROFILE=out ./myprogram
$ pprof --web ./myprogram out
'''

Forcing pid in all file names there will be super-annoying. I cannot agree to that.

I am truly sorry for saying yes to this and making you do the rebase thingy. Perhaps the idea with %p in file name is workable one. It will also let us deal with high bit thingy in better and cleaner way. Alternative is to look at extra variables like CPUPROFILE_USE_PID to signal that we need pids appended.

Please excuse me. And do let me know if you want to work on this larger change or not. In later case I am happy to make up for the trouble and author right change myself.

**Comment by artpol84** [comment]
> Forcing pid in all file names there will be super-annoying. I cannot agree to that.

It's ok I understand.

> Please excuse me. And do let me know if you want to work on this larger change or not. In later case I am happy to make up for the trouble and author right change myself.
NP

>  Alternative is to look at extra variables like CPUPROFILE_USE_PID to signal that we need pids appended.

This seems like the easiest way to do as long as you don't envision more items to be added to the filename. So I can implement that. 

In addition to that I'd propose to add the code that can detect HPC environments and adaptively do useful stuff there. For example
* Open MPI has OMPI_x, Slurm resource manager would have SLURM_x environment variables. MPICH is likely to have something similar (but need to check). We can use this as a sign that you are likely to want to have PID.
* In addition, if MPI is using PMIx (https://github.com/openpmix/openpmix) which is de-facto standard for Open MPI and is supported by MPICH, we can actually append MPI rank (instead of PID) to the output file name which will be even more convenient.
Let me know if those options are interesting for you.

**Comment by alk** [comment]
Sure. Having common mpi environments autodetected seems okay. Go ahead.

**Comment by artpol84** [comment]
@alk I've pushed the non-tested code to this PR. It builds but I've not checked it at runtime.
Could you give it a look and let me know if that works conceptually?

**Comment by alk** [comment]
Conceptually looks fine. I recommend adding more comments about motivation for detecting HPC environment (basically, what you wrote above likelyhood of same machine running multiple replicas of same program). If you don't mind, also consider adding comments on the high bit trick (why etc).

**Comment by alk** [comment]
LGTM. Lets get this in. Yes, please, squash and I'll merge.

**Comment by artpol84** [comment]
Squashed ... note that I haven't tested it with OMPI & Slurm yet

**Comment by alk** [comment]
Noted. You want to test it first? Or we assume if it doesn't work you/someone will fix it later ?

**Comment by alk** [comment]
I plan to do release in about week or so. So you don't need to rush.

**Comment by artpol84** [comment]
If you ok testing the common cases yourself and if they work I'll take care of testing the HPC case sometime later (in a week timeframe) and file a fix if it doesn't work.

**Comment by alk** [comment]
Yes I am fine testing basic case myself. So I still see 2 commits being proposed here. You still plan to squash them together? (I'd prefer that for sure)

**Comment by artpol84** [comment]
> So I still see 2 commits being proposed here. You still plan to squash them together? (I'd prefer that for sure)

Fixed!

File: src/base/sysinfo.cc [hunk 1]
```diff
@@ -206,6 +206,110 @@ extern "C" {
...
+
+  // No HPC environment was detected
+  return false;
+
+use_getpid:
+  snprintf(suffix, max_size, ".pid-%d", getpid());
+  return true;
+}
+
+void SetUniquePathSuffix(char *suffix, size_t max_size, bool is_child)
```
alk, line 285: So I see some duplication between here and above. Is it worth just having detect hpc thingy do the detect part and this function do the suffix part ?

src/base/sysinfo.cc
```diff
@@ -230,15 +334,20 @@ extern "C" {
 // TODO(csilvers): set an envvar instead when we can do it reliably.
 bool GetUniquePathFromEnv(const char* env_name, char* path) {
   char* envval = getenv(env_name);
+  bool is_child;
```
alk, line 337: it will be more readable to have this in same place as initialization below (and feel free to const it)

src/base/sysinfo.cc
```diff
@@ -206,6 +206,110 @@ extern "C" {
...
+{
+  enval[0] &= ~128;
+}
+
+static inline void EnvChildSignDrop(char *enval)
+{
+  enval[0] &= ~128;
+}
+
+static inline bool EnvChildSignIsOn(char *enval)
```
alk, line 233: inline is unnecessary. No huge deal but violates principle of least surprise

src/base/sysinfo.cc
alk: would be nice if you refactor so that no goto use_pid is necessary. Feels like it should be possible

src/base/sysinfo.cc
```diff
@@ -206,6 +206,110 @@ extern "C" {
...
+
+// HPC environment auto-detection
+// For HPC applications (MPI, OpenSHMEM, etc), it is typical for multiple
+// processes not engaged in parent-child relations to be executed on the
+// same host.
+// In order to enable gperftools to analyze them, these processes need to be
+// assigned individual file paths for the files being used.
+// The function below is trying to discover well-known HPC environments and
+// take advantage of that environment to generate meaningful profile filenames
+bool DetectHPCEnvironment(char *suffix, size_t max_size)
```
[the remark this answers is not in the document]
alk: So feel free to leave it to me. But we're passing pointer + size separately which isn't very amazing. In practice those are all pointers to char[PATH_MAX]. So perhaps we could be explicit there.

src/base/sysinfo.cc
```diff
@@ -230,15 +334,20 @@ extern "C" {
 // TODO(csilvers): set an envvar instead when we can do it reliably.
 bool GetUniquePathFromEnv(const char* env_name, char* path) {
```
alk, line 335: Somehow this commit broke unit tests.

Since the logic here becomes somewhat less obvious, can you please add unit test for GetUniquePathFromEnv ?

alk (changes requested):
Somehow it broke unittests. So needs revision

src/base/sysinfo.cc
```diff
@@ -206,6 +206,110 @@ extern "C" {
...
+// When the application is using fork() to spawn more processes, it is important
+// to avoid conflicts where both the parent and a child area using the same
+// filename.
+// In order to do so, the parent reads the user-defined environment and toggles
+// the MSB of the first byte in it's value to indicate that all other processes
+// using this variable are children (and thus) must append a unique identifier
+// (PID) as a suffix to the corresponsing filename.
+static inline void EnvChildSignSet(char *enval)
+{
+  enval[0] &= ~128;
```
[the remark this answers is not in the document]
artpol84: good catch! a copy/paste typo

src/base/sysinfo.cc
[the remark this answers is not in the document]
artpol84: yeah, I was thinking about that too. Will fix

src/base/sysinfo.cc
artpol84: That sounds like a good idea - we can test all the cases by just installing the right envar's!
I think the tests failed because of the bug in the child sign set

artpol84:
@alk please review.

along the way, I realized that MPI processes in rare cases may do fork so we want to keep adding PID in addition to HPC Rank in this case.

artpol84:
Unit tests is passing

alk:
tests are failing. For example heap-profiler_unittest

For the new file copyright year is not 2005 is not google. See for example safe_strerror.cc how it can be done. Also you sure that unique path test requires linking to tcmalloc_minimal ? Try minimal thing; only sysinfo and nothing else.

Why do you do fancy ptr and len thingy for AppendHPCEnvironment? It is always appending to empty PATH_MAX sized string. Why bother will all the extra logic? Also if you choose to keep append{Ptr,Len} then do make sure to keep naming style consistent.

artpol84:
> tests are failing. For example heap-profiler_unittest

Ok, I'll take a look. I was referring to my new test, haven't tried existing. 

> For the new file copyright year is not 2005 is not google. See for example safe_strerror.cc how it can be done.

Ok

> Also you sure that unique path test requires linking to tcmalloc_minimal ? Try minimal thing; only sysinfo and nothing else.

Ok, will try that

> Why do you do fancy ptr and len thingy for AppendHPCEnvironment? It is always appending to empty PATH_MAX sized string. Why bother will all the extra logic? 

It is not appending to the empty prefix always. At least this is not the case for the PID suffix. PID suffix may be on its own but also appended to the PMIX or Slurm suffixes already present.

> Also if you choose to keep append{Ptr,Len} then do make sure to keep naming style consistent.

Ok

src/tests/unique_path_unittest.cc
```diff
@@ -0,0 +1,295 @@
...
+  std::string expected = genDefaultParent();
+  expected += SLURM_SUFFIX;
+  expected += HPC_RANK;
+  return expected;
+}
+
+const std::string genSlurmChild()
+{
+  std::string expected = genSlurmParent();
+  appendPID(expected);
```
artpol84, line 125: @alk here for example, the child process in Slurm environment would have both Slurm and PID suffixes.

I decided to go ahead and always append to make sure that this is easily extendable, i.e. you can decide to add more suffixes when needed without breaking the logic.

artpol84:
> Also you sure that unique path test requires linking to tcmalloc_minimal ? Try minimal thing; only sysinfo and nothing else.

I wasn't sure about how to link with the whole LD_PRELOAD-able library. I can try adding only "sysinfo", but assumed it might have dependencies either now or in the future that would introduce additional maintenance hassle later. 
IMO it's a good practice to link the test to the whole library. In this case, you always have all dependencies correct

artpol84:
For me the heap profiler is passing:

```
$ git log -1 && ./heap-profiler_unittest 
commit bd6a6ea2be62e6a456fd4d83c4b18cf30ced353a (HEAD -> topic/fix_outfile_v2, origin/topic/fix_outfile)
Author: Artem Polyakov <artpol84@gmail.com>
Date:   Thu Jul 13 17:43:29 2023 -0700

    Add unit test for GetUniquePathFromEnv()
    
    Signed-off-by: Artem Polyakov <artpol84@gmail.com>
Starting tracking the heap
Starting tracking the heap
DONE.
```

Am I looking at the wrong place? Should I run it in some special way?

artpol84:
> Also if you choose to keep append{Ptr,Len} then do make sure to keep naming style consistent.

Will `AppendPtr()` and `AppendLen()` be consistent with the naming style?

artpol84:
Ok, I guess I was able to reproduce
```
$ ./heap-profiler_unittest.sh
Profile not found: /tmp/heap-profiler_unittest.beLeuR/test_*.1329.heap
```

Fixing now

artpol84:
Tests should be passing now.
```
$ ./heap-profiler_debug_unittest.sh && ./heap-profiler_unittest.sh 
PASS
PASS
```

I also fixed the naming of the functions per below:
> Will `AppendPtr()` and `AppendLen()` be consistent with the naming style?

artpol84:
@alk 
* I've fixed the file header to be "Copyright (c) 2023, gperftools Contributors"
* I tried building just sysinfo.cc into the test but was getting some errors at the 'autogen.sh' running stage (see below). What I did is I removed deps from tcmalloc and simply put "sysinfo.cc" to the _SORCES variable in Makefile.am.

```
$ ./autogen.sh
Makefile.am: error: object 'src/base/sysinfo.$(OBJEXT)' created both with libtool and without
autoreconf: automake failed with exit status: 1
```

alk:
Thanks for update. Several more questions. (Some I can fix myself but this is never ideal if I have to amend people's work)

* I see that you added makefile reference to unique_path_unittest.vcxproj but there is no such file. It is not great that we're not testing this on windows, but is better than breaking make dist command.
* CheckPIDIsForced should be static
* using PATH_MAX-sized buffer for environment name string is mildly confusing. Breaks principle of least surprise. I am not sure how to size this one but PATH_MAX we should probably avoid.
* Slurm's docs mention that SLURM_JOBID is deprecated. Also can you clarify why we need this approach "if jobid is there but not procid then we force pid" ?
* For testing checking XYZ_USE_PID thing, please use EnvToBool
* and most general, this is adding a lot of abstractions. Most used exactly once. And many of them are just giving names to obvious things and result is big diff with logic spread quite thinly. I understand it is matter of taste, but we don't do that kind of style in gperftools. Bigger offenders here are stuff like PID_SUFFIX but those numerous functions are questionable as well.
* I think general logic can be simplified with hpc detection bits (which don't need to be own function) set up something like mpi_sep and mpi_val. And then you format it all at once something like this:

```
  if (append_pid) {
    snprintf(path, PATH_MAX, "%s%s%s_%d", envval, mpi_sep, mpi_val, getpid());
  } else {
    snprintf(path, PATH_MAX, "%s%s%s", envval, mpi_sep, mpi_val);
  }
```

which saves all the trouble with AppendXYZ stuff (and confusion inspiring O(N^2) behaviors; yes, I might by only one irked by this and this code isn't exactly speed critical but still)

alk:
Ah and yes, fine to leave makefile bits to me. Dependency on libtcmalloc is unfortunate and fixable, but several other unit tests are in same shape. So I'll be fixing that separately. And I can't be asking people to learn autotools bits.

artpol84:
I see you closed it. What are the next steps? From responses it seems you were asking for further fixes, but in this case it should not be closed right?

alk:
Oh. I apologize. Had no intention to close. Must have hit by mistake. Yes, if you could adjust further I'd be thankful.

artpol84:
I will fix as requested

artpol84:
@alk, I did a pass to address the comments. Please see below

> * I see that you added makefile reference to unique_path_unittest.vcxproj but there is no such file. It is not great that we're not testing this on windows, but is better than breaking make dist command.

Fixed - I've removed that item from the Makefile.am as I'm not sure how to properly generate it.

> * CheckPIDIsForced should be static

N/A - I've got rid of this function per one of the requests below

> * using PATH_MAX-sized buffer for environment name string is mildly confusing. Breaks principle of least surprise. I am not sure how to size this one but PATH_MAX we should probably avoid.

Not sure what's the best way to handle it. 
I've introduced the following macro's in the sysinfo.h:
```C
#define TC_ENV_MAX_NAME 256
```

that defines the max expected variable name.

> * Slurm's docs mention that SLURM_JOBID is deprecated. Also can you clarify why we need this approach "if jobid is there but not procid then we force pid" ?

Good catch, fixed

> * For testing checking XYZ_USE_PID thing, please use EnvToBool

Fixed

> * and most general, this is adding a lot of abstractions. Most used exactly once. And many of them are just giving names to obvious things and result is big diff with logic spread quite thinly. I understand it is matter of taste, but we don't do that kind of style in gperftools. Bigger offenders here are stuff like PID_SUFFIX but those numerous functions are questionable as well.

Indeed, this is very subjective. I tried to address, but I'm not sure I am able to guess what you would prefer:
1. I turned Child bit manipulation into macro's
1. I've merged `CheckPIDIsForced()` into `GetUniquePathFromEnv()`
1. As of now, only 1 new function (`QueryHPCEnvironment()`) introduced that I'd prefer to keep separate from both readability and logic standpoints.



> * I think general logic can be simplified with hpc detection bits (which don't need to be own function) set up something like mpi_sep and mpi_val. And then you format it all at once something like this:
> 
> ```
>   if (append_pid) {
>     snprintf(path, PATH_MAX, "%s%s%s_%d", envval, mpi_sep, mpi_val, getpid());
>   } else {
>     snprintf(path, PATH_MAX, "%s%s%s", envval, mpi_sep, mpi_val);
>   }
> ```
> 
> which saves all the trouble with AppendXYZ stuff (and confusion inspiring O(N^2) behaviors; yes, I might by only one irked by this and this code isn't exactly speed critical but still)

Fixed

I've also added some macros to sysinfo.h to make sure that the unit test and sysinfo.cc are aligned. I know you didn't want to use the macro for this, but I believe it's more reliable as we actually using these things from 2 places.
If you still think we need to get rid of this - it's fine, I can copy/paste in all the places. But IMO this makes sense for easier maintenance.

```C
#define TC_ENV_PID_SUFFIX "_"
#define TC_ENV_PMIX_RANK "PMIX_RANK"
#define TC_ENV_PMIX_SUFFIX ".rank-"
#define TC_ENV_SLURM_SUFFIX ".slurmid-"
#define TC_ENV_SLURM_JOBID "SLURM_JOB_ID"
#define TC_ENV_SLURM_PROCID "SLURM_PROCID"
```

alk:
applied. thanks. (I ended up massaging your changes in separate commit)

artpol84:
Thanks!
"massaging" looks great - I do not think in terms of OOP, unfortunately. So I doubt I'd come with up something close to that one on my own.

PR #1264, opened 2021-04-03
tcmalloc page fences: add TCMALLOC_PAGE_FENCE_READABLE option

risicle:
This is a hack I added a while back and never thought to upstream. I found it useful for sifting out OOB writes from less dangerous OOB reads.


risicle:
Um, I have no intention of getting a google account.

alk:
Actually, I get this merged regardless of googlebot thingy. I am not quite certain how useful it is, but it is a small and simple change so why not.

alk:
Applied. Thanks.

PR #1269, opened 2021-04-15
Conditionally set default_tcmalloc_pagesize to kernel pagesize

rosinL:
set default_tcmalloc_pagesize to kernel pagesize when default_tcmalloc_pagesize
is smaller than kernel pagesize.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -104,6 +104,11 @@ if test "$enable_minimal" = yes; then
   enable_heap_profiler=no
   enable_heap_checker=no
 fi
+pagesize=$(getconf PAGESIZE 2>&1)
+tcmalloc_pagesize_byte=$(expr $default_tcmalloc_pagesize \* 1024)
+if test -n "$pagesize" && test "$pagesize" -gt "$tcmalloc_pagesize_byte"; then
+  default_tcmalloc_pagesize=$(expr $pagesize / 1024)
+fi
 AC_ARG_ENABLE([stacktrace-via-backtrace],
               [AS_HELP_STRING([--enable-stacktrace-via-backtrace],
                               [enable use of backtrace() for stacktrace capturing (may deadlock)])],


rosinL:
pls help to review @alk, thanks

kevinzs2048:
@alk Could you help to review? Really appreciated!

alk:
Apologies for the delay and thanks for the ping.

Unfortunately, I dont think this is right way of fixing the situation with larger page sizes.

My understanding is there is no any special "64k pages ABI" on arm (or riscv). Instead, all binaries are expected to run in both 4k environments and 64k pages environments. We can see that by observing that various sections in the binary (e.g. .data) are aligned on 64k on arms (and 4k on x86-es).

So proper fix for this problem is making sure that gperftools behaves correctly whenever larger page size is in effect at runtime. One way of doing it would be to just have arms always do 64k tcmalloc pages, but that is too large at least for some environments. So better way is to "just" have different notions of page size when it comes to page map (this size has to be hardcoded in the binary as it is used in fast-path of deallocation), and of page size when it comes to various mm syscalls. It will probably not be a small change but it is a right thing to do.

Until this is done, I just don't get why people that are heavily affected by this issue are not manually configuring 64k pages via ./configure.

kevinzs2048:
Hi @alk, thanks for your review. 
After some investigation around different distro and architecture, we found that the page size varies based on distro, not on specified architecture. For example, on Debina/Ubuntu/SUSE, the page size is set as 4K, and in Centos, the pagesize is 64K, no matter what architecture they have. 

So that our proposal is to align the tcmalloc default pagesize with the OS distro page size. The alignment can farthest keep the pace with the OS. Please let me know your thoughts, thanks a lot!

alk:
Thanks. So let me just confirm. Are you saying that rhel/centos on arm64 will refuse to run 4k binaries ? Or will, say, glibc on those distros refuse to run 4k pages kernels ?

PR #1270, opened 2021-04-20
README: replace "golang" moniker with "Go"

matttproud:
This commit replaces the "golang" moniker in the README file with the
appropriate name of "Go" per https://golang.org/doc/faq#go_or_golang.

In this context, no disambiguation is required.


alk:
Applied. Thanks a lot.

PR #1290, opened 2021-07-30
Add an arm64 implementation for SpinlockPause()

AGSaidi:
This patch adds an arm64 implementation of the SpinlockPause() function
allowing the core to adaptively spin to acquire the lock and improving
performance in multi-threaded environments where the locks can be contended.

From experience with other projects, we've found a single isb is the closest
corollary to the rep; nop timing or x86.

Overall, across thirty runs, the binary_trees_shared benchmark improves 6% in
average, median, and P90 runtime with this change.


AGSaidi:
Any feedback?

snadampal:
Any chance of getting this merged soon? this is going to help multiple projects.

snadampal:
@alk , would you be able to review and merge this? thank you!

alk:
Thanks. Applied.

PR #1299, opened 2021-09-14
free memory with mmap(PROT_NONE, MAP_FIXED)

vnikolai:
For some platforms that madvise(ptr, size, MADV_FREE) has no effect here is alternative solution provided with mmap(PROT_NONE, MAP_FIXED) to keep pages mapped but not backed up by physical memory pages. That counts as free by some systems/tools, tested on QNX with platform available tools.


MashoujiangPlusAI:
Dear, @vnikolai , looks like you have porting tcmalloc on QNX platform? could you please provide more info about how to porting tcmalloc? really appreciate!

PR #1300, opened 2021-09-14
free memory with mmap(PROT_NONE, MAP_FIXED)

vnikolai:
For platforms that does not honor madvise and it has no effect or unclear effect propose alternative solution with mmap(PROT_NONE, MAP_FIXED) to keep pages mapped but not backed up by physical memory pages. Depending on underlying behavior it might have better effect.

### DIFF
--- a/src/system-alloc.cc
+++ b/src/system-alloc.cc
@@ -511,7 +511,7 @@ void* TCMalloc_SystemAlloc(size_t size, size_t *actual_size,
 }
 
 bool TCMalloc_SystemRelease(void* start, size_t length) {
-#ifdef MADV_FREE
+#if defined(FREE_MMAP_PROT_NONE) && defined(HAVE_MMAP) || defined(MADV_FREE)
   if (FLAGS_malloc_devmem_start) {
     // It's not safe to use MADV_FREE/MADV_DONTNEED if we've been
     // mapping /dev/mem for heap memory.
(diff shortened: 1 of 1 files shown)


src/system-alloc.cc
```diff
@@ -545,11 +554,19 @@ bool TCMalloc_SystemRelease(void* start, size_t length) {
...
   return false;
 }
 
 void TCMalloc_SystemCommit(void* start, size_t length) {
+#if defined(FREE_MMAP_PROT_NONE) && defined(HAVE_MMAP)
+  // remaping as MAP_FIXED to same address assuming span size did not change 
+  // since last TCMalloc_SystemRelease
+  mmap(start, length, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_ANONYMOUS|MAP_FIXED,
+       -1, 0);
+#elif defined(MADV_FREE)
```
alk, line 567: one litttle nit-pick here, why bother with elif if you can simply do else ?

alk:
This is really nice. One little ask about replacing (in 2 place) elsif with MADV_FREE to just else which would seem more logical. Also, since madvise code handles EINTR (but yes AFAIK linux doesn't return it for mm syscalls), so might be worth doing same just for principle of least surprise.

Also thanks for making it optional. Indeed, your proposal works, but on linux both "release" and "commit" calls will take mmap_sem lock for write, while madvise (both dontneed and free) are doing it for read, and page faults which occur after empty "commit" calls also do read. So while it will fix some people's use-cases, unfortunately it cannot be default.

src/system-alloc.cc
vnikolai: Done

vnikolai:
@alk did you mean handle EAGAIN same manner as madvise? EINTR confuses me here

vnikolai:
@alk Kindly reminder please review updated version.

alk:
Applied. Thanks. Can you clarify what OSes this is intended for ? (it works slightly less great on Linux locking-wise and  I suspect many other unix-like OSes would have same outcomes)

vnikolai:
That is designed for QNX. And tested also by confirming memory consumption with hogs utility.

Issue #1302, opened 2021-09-24
a way to avoid null pointer dereference at no Debug build


kashirin-alex:
_The noticeable build warning at LTO (while tcmalloc is built with `-flto -fuse-linker-plugin -ffat-lto-objects` and `-O3` in this case):_
```CPP
/root/s/gperftools/src/central_freelist.cc: In member function ‘ReleaseToSpans’:
/root/s/gperftools/src/central_freelist.cc:110:13: warning: potential null pointer dereference [-Wnull-dereference]
  110 |   if (span->objects == NULL) {
```

* That is only a way to avoid a segmentation (suppose nullptr has happen at no-debug) and to observe the warning: potential null pointer dereference -- It does not resolve a case when or why the object is out of pages map. 

as the object is for release, as like the behavior of return on nullptr at `void CentralFreeList::ReleaseListToSpans(void* start)` https://github.com/gperftools/gperftools/blob/c25941200ef4ce39d0774c1332ff7abfbeab7035/src/central_freelist.cc#L83 

The `void CentralFreeList::ReleaseToSpans(void* object)` https://github.com/gperftools/gperftools/blob/c25941200ef4ce39d0774c1332ff7abfbeab7035/src/central_freelist.cc#L104
can return on `span` is nullptr as the `Static::pageheap()->GetDescriptor(p);` is nullptr when nothing to release (out-of-range)
https://github.com/gperftools/gperftools/blob/f7c6fb6c8e99d6b1b725e5994373bcd19ffdf8fd/src/pagemap.h#L146

> A finding, the `Span* MapObjectToSpan(void* object)` for not GCC-4.5.0 can be considered for `inline ATTRIBUTE_ALWAYS_INLINE`  (competes the call-graph inlinement instructions)

kashirin-alex:
Further more, the `Span* span` evaluation  
 https://github.com/gperftools/gperftools/blob/c25941200ef4ce39d0774c1332ff7abfbeab7035/src/central_freelist.cc#L105
should/can be before `SLL_Next` by that guarantees Range is good. (and will have a loop-exit in case the Span is a out-of-range/bad address)

-- Are the facts on the Issue good enough for a pull-request ? (just pre-checking)

alk:
[...]

PR #1305, opened 2021-09-29
check Span* before Next & ReleaseToSpans

kashirin-alex:
added Span* check  at `void CentralFreeList::ReleaseListToSpans(void* start)` before calling/using the start-pointer.
changed `void CentralFreeList::ReleaseToSpans` added Span* argument `(void* object, Span* span)`
added Macro `LOG_LEVEL_EXPECT` at Debug "Crash" (as ASSERT) rest only "Log" information.

The difference in the result, In a non-debug build instead of a segmentation on nullptr access, the Bad Span will be logged and break the release loop. A debug build (as use to be) will abort with additional Bad Span information.
- In production run-time (instead the Segmentation at nullptr), There is the need, at Log-Info found, of an app restart and report the logs cases of `Bad Span`/`Bad List` .

Fixes: #1302

### DIFF
--- a/src/central_freelist.cc
+++ b/src/central_freelist.cc
@@ -79,14 +79,6 @@ void CentralFreeList::Init(size_t cl) {
   ASSERT(cache_size_ <= max_cache_size_);
 }
 
-void CentralFreeList::ReleaseListToSpans(void* start) {
-  while (start) {
-    void *next = SLL_Next(start);
-    ReleaseToSpans(start);
-    start = next;
-  }
-}
-
 // MapObjectToSpan should logically be part of ReleaseToSpans.  But
 // this triggers an optimization bug in gcc 4.5.0.  Moving to a
 // separate function, and making sure that function isn't inlined,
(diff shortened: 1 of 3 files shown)


kashirin-alex:
@googlebot I signed it!

alk:
Thanks for the patch. But is it really useful for broken process to deliberately avoid crashing ? Can you elaborate why this is done ?

kashirin-alex:
The main reason, It is to have the possibility for a grace-full shutdown, that is for cases such as commit data to files.
I can consider a good choice for quitting at broken-process with `std::raise(SIGTERM);` after the info logging.
That is a perfect form for the SWC-DB project; The applications will perform a shutdown sequence https://github.com/kashirin-alex/swc-db/blob/9e374a0cf269cabf71185d02ee4928466d7996a0/src/cc/include/swcdb/ranger/AppContext.h#L257. 
If such app-exit with `std::raise(SIGTERM);` is desired/compatible, I'll add a commit to this pull-request.

kashirin-alex:
to extend, if to use the `std::raise(SIGINT)` it should be going as last check of mem-release return and on fail to signal, which should be replacing all the asserts with an error-state (0:true, 1:false, 2: failure).
Just let me know, if `std::raise(SIGINT)` can be considered for the duty. Which again in this case is only for the free/release memory where the program can continue to run.

if the use of `std::raise(SIGINT)` is to complicate to decide upon, the most useful will be to add static function ptr hook for mem-release failure and in this case the default will be the assert as by the `ifndef NDEBUG` .
-- I haven't in-counted with release ASSERT problems, whereas it is possible to apply a strong(observable) process for tolerance to failure.

alk:
I am sorry for getting back to you so late. So, unfortunately I have to reject this. "Handling" broken freelist by ignoring bad objects is against general philosophy of crashing fast (if/break statements that you added instead of ASSERTs)

Issue #1312, opened 2021-10-29
tcmalloc incompatible with macOS Monterey?


ravenAtSafe:
Previous to Monterey tcmalloc was incompatible with apps launched from the Dock.  As of Monterey tcmalloc causes apps to crash even when launched from the command line.  I've tried 2.7 and 2.9.1.

See also https://trac.macports.org/ticket/63242 which I assume is the same issue.

The resulting stack is pretty useless as far as I can tell, but here you go.

```
Process 56526 stopped
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x0)
    frame #0: 0x0000000000000000
error: memory read failed for 0x0
Target 0: (myapp) stopped.
(lldb) thread backtrace
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x0)
  * frame #0: 0x0000000000000000
    frame #1: 0x00007ff818508bb2 libobjc.A.dylib`load_images + 1358
    frame #2: 0x000000010006341c dyld`dyld4::RuntimeState::notifyObjCInit(dyld4::Loader const*) + 170
    frame #3: 0x0000000100068bfd dyld`dyld4::Loader::runInitializersBottomUp(dyld4::RuntimeState&, dyld3::Array<dyld4::Loader const*>&) const + 167
    frame #4: 0x0000000100068beb dyld`dyld4::Loader::runInitializersBottomUp(dyld4::RuntimeState&, dyld3::Array<dyld4::Loader const*>&) const + 149
    frame #5: 0x0000000100068beb dyld`dyld4::Loader::runInitializersBottomUp(dyld4::RuntimeState&, dyld3::Array<dyld4::Loader const*>&) const + 149
    frame #6: 0x0000000100068beb dyld`dyld4::Loader::runInitializersBottomUp(dyld4::RuntimeState&, dyld3::Array<dyld4::Loader const*>&) const + 149
    frame #7: 0x0000000100068cac dyld`dyld4::Loader::runInitializersBottomUpPlusUpwardLinks(dyld4::RuntimeState&) const + 108
    frame #8: 0x000000010007c32e dyld`dyld4::APIs::runAllInitializersForMain() + 222
    frame #9: 0x000000010005a358 dyld`dyld4::prepare(dyld4::APIs&, dyld3::MachOAnalyzer const*) + 3438
    frame #10: 0x00000001000594b4 dyld`start + 388
[...]

PR #1315, opened 2021-11-18
Implement and register free_definite_size for macOs

natale-p:
In Monterey, it seems that free_definite_size is required (despite the documentation saying that is optional). The implementation just forwards the call to free.

Fix: #1312 

Signed-off-by: Natale Patriciello <natale.patriciello@gmail.com>


natale-p:
@googlebot I signed it!

alk:
Applied. Thanks.

PR #1316, opened 2021-11-21
Detect Valgrind based on LD_PRELOAD patterns.

yurivict:
This patch adds a more reliable way to detect that the process is run under Valgrind.

It is based on this suggestion: https://stackoverflow.com/questions/365458/how-can-i-detect-if-a-program-is-running-from-within-valgrind

### DIFF
--- a/src/base/dynamic_annotations.c
+++ b/src/base/dynamic_annotations.c
@@ -43,6 +43,19 @@
 #include "base/dynamic_annotations.h"
 #include "getenv_safe.h" // for TCMallocGetenvSafe
 
+static int running_on_valgrind_preload = -1;
+void __attribute__ ((constructor)) premain() {
+  char *LD_PRELOAD = getenv("LD_PRELOAD");
+  running_on_valgrind_preload = LD_PRELOAD != NULL &&
+    (
+      strstr(LD_PRELOAD, "/valgrind/") != NULL
+      ||
+      strstr(LD_PRELOAD, "/vgpreload") != NULL
+    )
+    ?
+    1 : 0;
+}
+
 static int GetRunningOnValgrind(void) {
 #ifdef RUNNING_ON_VALGRIND
   if (RUNNING_ON_VALGRIND) return 1;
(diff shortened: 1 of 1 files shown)


src/base/dynamic_annotations.c
```diff
@@ -43,6 +43,19 @@
 #include "base/dynamic_annotations.h"
 #include "getenv_safe.h" // for TCMallocGetenvSafe
 
+static int running_on_valgrind_preload = -1;
+void __attribute__ ((constructor)) premain() {
```
alk, line 47: this will not work on non-gcc/non-clang compilers (and intel on gnu/linux also supports those).

Can you elaborate what is the issue you're fixing or improving ?

yurivict: > Can you elaborate what is the issue you're fixing or improving ?

Currently gperftools doesn't detect by itself that it runs under Valgrind. Detection is based on the compiler define value ```RUNNING_ON_VALGRIND``` which isn't set automatically, and alternatively on the environment variable ```RUNNING_ON_VALGRIND``` which can be set by the user. Both aren't automatic.

This patch allows gperftools to auto-detect Valgrind on Linix-like systems.

I maintain the gperftools port on FreeBSD. The user contacted me that he was in pain because there is no auto-detection of Valgrind in gperftools. I added this patch to the port, which satisfied the user.

This solution's idea is taken from [here](https://stackoverflow.com/questions/365458/how-can-i-detect-if-a-program-is-running-from-within-valgrind).

alk: I see. Yes, this seems justified. So just implementation needs to be different. First, existing code that runs (apparently insufficient) checks for valgrind is running early enough. No need to have extra "premain" thingy doing it.

Second, getenv sometimes works, but sometimes it is too early for it to work so early during first call to malloc. So we have special function just for that, inspecting environment early in program startup. There is GetenvBeforeMain.

Also maybe, but not sure, I'd consider to maybe only check LD_PRELOAD on unix-y platforms. Or maybe not worth any complexity to add those more ifdefs.

alk:
Gentle ping. You still want some version of this merged?

PR #1320, opened 2021-12-27
Continue to release span until the end of one round.

liqingqing1229:
There have so much more situations which will release span failed. such as: when default_tcmalloc_pagesize is smaller than kernel pagesize(#1269) , and this will lead to failure to recycle resource in time, and cause memory growth(#1201).
So continue to release span until the end of this round. And this is also sync from the google/tcmalloc project.

### DIFF
--- a/src/page_heap.cc
+++ b/src/page_heap.cc
@@ -483,11 +483,18 @@ Length PageHeap::ReleaseSpan(Span* s) {
 
 Length PageHeap::ReleaseAtLeastNPages(Length num_pages) {
   Length released_pages = 0;
+  Length prev_released_pages = Length::max() + Length(1);
 
   // Round robin through the lists of free spans, releasing a
   // span from each list.  Stop after releasing at least num_pages
   // or when there is nothing more to release.
   while (released_pages < num_pages && stats_.free_bytes > 0) {
+    if (released_pages == prev_released_pages) {
+      // Last iteration of while loop made no progress.
+      break;
+    }
+    prev_released_pages = released_pages;
+
     for (int i = 0; i < kMaxPages+1 && released_pages < num_pages;
          i++, release_index_++) {
       Span *s;
@@ -510,8 +517,6 @@ Length PageHeap::ReleaseAtLeastNPages(Length num_pages) {
       // large freelist, should we carve s instead of releasing?
       // the whole thing?
       Length released_len = ReleaseSpan(s);
-      // Some systems do not support release
-      if (released_len == 0) return released_pages;
       released_pages += released_len;
     }
   }


alk:
Thanks for the patch, but what do you think about https://github.com/alk/gperftools/pull/new/arm-64k-pages instead ?

liqingqing1229:
> Thanks for the patch, but what do you think about https://github.com/alk/gperftools/compare/arm-64k-pages?expand=1 instead ?

Hi alk, thanks for your reply. I have read the arm-64k-pages branch's source code and yes this can resolve the problem when the system page size large than kPagesize, and this is similar like this pr: https://github.com/gperftools/gperftools/pull/1269.

But i think the issue is that the basic memory overhead increases cause most memory applications are small memory.
for example we allocate to object which are 64B and  128B. before that, we just need 64KB physic addr, but after that if we use a 64KB for a basic granularity, this maybe use 128KB.
so what i think is that we just release the span timely.

thanks.

kashirin-alex:
as for "release the span timely.", Not sure, but Most likely you want ( prefer) to achieve a ` ::malloc_trim(keep)` functionality (example https://github.com/kashirin-alex/swc-db/blob/49351ab964441f7d4cfa64e4fecb72abe0bff30b/src/cc/include/swcdb/common/sys/Resource_Mem.h#L429) 
 
That is in place of Zero `stats_.free_bytes > 0` it is to 'keep n bytes' in the released_pages

(malloc_trim ain't a solution to the pagesize differences ) while more problematic cause is why would `PageHeap::ReleaseAtLeastNPages(Length num_pages)` be called with more pages than possible to release

liqingqing1229:
> as for "release the span timely.", Not sure, but Most likely you want ( prefer) to achieve a ` ::malloc_trim(keep)` functionality (example https://github.com/kashirin-alex/swc-db/blob/49351ab964441f7d4cfa64e4fecb72abe0bff30b/src/cc/include/swcdb/common/sys/Resource_Mem.h#L429)
> 
> That is in place of Zero `stats_.free_bytes > 0` it is to 'keep n bytes' in the released_pages
> 
> (malloc_trim ain't a solution to the pagesize differences ) while more problematic cause is why would `PageHeap::ReleaseAtLeastNPages(Length num_pages)` be called with more pages than possible to release

Actually,  what i want is to reduce system memory usage when the scenario is that: the system pagesize large than kPagesize. ie: when I use tcmalloc in centos8 in aarch64, the system pagesize is 64k, and the default tcmalloc's kPagesize is 8k.  As I said,  kPageSize = 8k is really ok for me, I don't want to use larger kPagesize to manage the memory cause it's a little waste memory. 

In this situation,  the tcmalloc api  call  PageHeap::ReleaseAtLeastNPages and this function will call ReleaseSpan,  this is easy to return failed. When this happend, current PageHeap::ReleaseAtLeastNPages will return immediately, and the  "release_index_" won't be updated, the consequence is that next time when we call PageHeap::ReleaseAtLeastNPages again,  it's quite possibly to return failed immediately cause we use the same release_index_ to release span and tcmalloc will alloc more system memory instead.  
But if we continue to release span until the end of one round, this will do more MergeIntoFreeList in PageHeap::ReleaseSpan, and when we free pages to system, This will easier to succeed.

liqingqing1229:
sorry, this pr will cause compile error

alk:
Indeed, the downside of my approach is that it will slightly increase fragmentation as we'll use larger spans for allocations. So it will be as if we compiled malloc with 64k logical pages. I.e. it will be less likely for span of say 32 bytes size class to get fully empty in 64k spans than 8k spans.

But in practice increase in memory fragmentation from larger spans for most workloads isn't large.

On the other hand "continue to try" approach you're advocating doesn't guarantee successful return of pages to kernel, as you need to locate a span which start and end address aligns exactly on 64k boundary. Chance of this occurring by luck isn't large. And the cost of such trying is more overhead particularly in environments where release cannot work. E.g. when someone configured 1GB huge page backing of heap.

Indeed, in perfect world, we would handle it like we should handle huge pages situation. We would still allow single 64k OS page to have arbitrary subset of 8k spans (or even 4k spans). We would somehow implement great heuristics such that we actually minimize number of used OS pages (and so have real chance of getting fully free 64kb chunks to release), and we would have ways to find those releasable chunks quickly (i.e. much faster than O(N); i.e. not "keep trying" for sure).

Perhaps one approach could be to try to apply 'hugepages aware' page heap replacement from google tcmalloc (github.com/google/tcmalloc). But while it implements some heuristics that appear to help in "packing" memory usage in fewer huge pages, it has tons of issues and in my view doesn't justify huge increase in complexity. But something along this lines, getting our page heap friendlier for huge page support, would be best outcome for this case too.

PR #1329, opened 2022-03-21
avoid linux uapi headers

Hello71:
this avoids needing to install linux-headers to build gperftools.


alk:
Thanks for the patch, but I already eliminated linux-syscall-support thingy. I still see one include of linux/ptrace so will right now check if this can be replaced as per your patch. Thanks for bringing this up.

alk:
Thanks. Merged your suggestion at https://github.com/gperftools/gperftools/commit/fe5ba4b5242c143661989a15a981079957835c44

Issue #1333, opened 2022-04-21
SpinLockDelay in spinlock_win32-inl.h could be rewritten with WaitOnAddress for better performance


lennoxho:
This snippet in spinlock_win32-inl.h
https://github.com/gperftools/gperftools/blob/fe85bbdf4cb891a67a8e2109c1c22a33aa958c7e/src/base/spinlock_win32-inl.h#L41-L51
could be rewritten with [WaitOnAddress](https://docs.microsoft.com/en-us/windows/win32/api/synchapi/nf-synchapi-waitonaddress)
WaitOnAddress is equvalent to futex, which is the solution used on Linux.

```C++
void SpinLockDelay(volatile Atomic32 *w, int32 value, int loop) {
    if (loop != 0) {
        auto wait_ns = static_cast<uint64_t>(base::internal::SuggestedDelayNS(loop)) * 16ull;
        auto wait_ms = wait_ns / 1000000ull;

        WaitOnAddress(w, &value, 4u, static_cast<DWORD>(wait_ms));
    }
}

void SpinLockWake(volatile Atomic32 *w, bool all) {
    if (all) WakeByAddressAll((void*)w);
    else WakeByAddressSingle((void*)w);
}
```
In our project, we have 10 threads performing small allocations in parallel.
For this workload, we are seeing massive (10%-ish) improvement in runtime.

This VTune chart perfectly demonstrates what the change does to CPU utilisation
![image](https://user-images.githubusercontent.com/9140100/164534102-384ea964-cf14-431c-964c-3b6a97f95b19.png)
(top: before; bottom: after)

alk:
Yes. Patches welcome. Also consider "stealing" whatever is being done in abseil. They may have modernized this too.

shiyi23:
@alk could you asign me？
/asign

alk:
Oh, wasn't even aware there is such feature. Sure, if you ask.

lennoxho:
New PR attempt which addresses the comment to support Windows 7 or older.
https://github.com/gperftools/gperftools/pull/1438

lennoxho:
New PR https://github.com/gperftools/gperftools/pull/1440

PR #1350, opened 2022-06-28
code refactor#rewritten with WaitOnAddress for better performance.

shiyi23:
issue id: [1333](https://github.com/gperftools/gperftools/issues/1333)

### DIFF
--- a/src/base/spinlock_win32-inl.h
+++ b/src/base/spinlock_win32-inl.h
@@ -34,6 +34,8 @@
 
 
 #include <windows.h>
+#include <synchapi.h>
+#include <cstdint>
 
 namespace base {
 namespace internal {


alk:
Thanks a lot! Looking. Do we need to "#include <synchapi.h>" ?

alk:
Hm, the docs mention that it requires at least windows 8. And there are still some users of windows 7. So worth a think.

[1] https://docs.microsoft.com/en-us/windows/win32/api/synchapi/nf-synchapi-waitonaddress

[2] https://www.statista.com/chart/2322/market-share-of-desktop-operating-systems/#:~:text=According%20to%20estimates%20from%20Statcounter,points%20from%20a%20year%20earlier.

alk:
Looking at what chrome does, I think we should keep support for win7. Can you do ifdef or something (getprocaddr perhaps to dynamically "probe" for waitonaddress ?)

shiyi23:
> Looking at what chrome does, I think we should keep support for win7. Can you do ifdef or something (getprocaddr perhaps to dynamically "probe" for waitonaddress ?)

Thank you for your valuable comments! I'll keep optimizing this code!

lennoxho:
> Looking at what chrome does, I think we should keep support for win7. Can you do ifdef or something (getprocaddr perhaps to dynamically "probe" for waitonaddress ?)

Make sure to do this binding during tcmalloc startup procedures to avoid the cost of checking each time SpinLockDelay is called.

lennoxho:
Another thing we might want to consider is the performance on Windows 10 (ARM64).

lennoxho:
Thanks for looking into this.
Sorry I didn't follow up on #1333. My company has a tedious FOSS contribution process.

shiyi23:
> Thanks for looking into this. Sorry I didn't follow up on #1333. My company has a tedious FOSS contribution process.



> Thanks for looking into this. Sorry I didn't follow up on #1333. My company has a tedious FOSS contribution process.

No matter, I thank you very much, your work inspires me!

lennoxho:
New PR attempt which addresses the comment to support Windows 7 or older.
https://github.com/gperftools/gperftools/pull/1438

alk:
Thanks for the patch. We took the new proposed patch.

Let me also clarify what happened there. Since in this pull request I argued that we need to support win 7, while what we ended up taking is win8+ only.

The idea is _back then_ when this pull request was proposed, win7 was seemingly still supported by bunch of software as noted above. But today, win7 is pretty much abandoned by all major software packages. Thus decision to simplify.

Issue #1375, opened 2022-12-07
pprof: request: exclude allocations that have since been freed


Spongman:
So, in pprof there's a `--base` argument to exclude allocations that were not made until after a certain point.

I'd like to request a similar argument that allows us to exclude allocations that were not live _after_ a given point.

For example, if I have 3 heap files: `0001.heap`, `0002.heap` and `0003.heap`, I only want to show the memory that was allocated when `0002.heap` was dumped, and _specifically exclude_ those allocations that were freed before `0003.heap` was allocated.

alk:
I am sorry but we're not supporting/improving our perl version of pprof.

PR #1376, opened 2022-12-07
Update pprof

Spongman:
allow multiple `:`-separated profiles to be subtracted from the main profile.

### DIFF
--- a/src/pprof
+++ b/src/pprof
@@ -622,12 +622,15 @@ sub Main() {
     }
   }
 
-  # Subtract base from profile, if specified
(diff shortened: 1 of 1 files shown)


Spongman:
fixes #1375

alk:
Hi and thanks for the patch. I am sorry but we're not supporting our own perl version of pprof. Please use github.com/google/pprof.

As for how to subtract multiple profiles, (with google's proper pprof) you can first add several profiles by using --proto output. And then that sum you can subtract from whatever you want by using usual --base option. (Or consider contributing this special feature there)

Closing as won't fix. Sorry.

PR #1379, opened 2022-12-26
Define off64_t as off_t on musl

kraj:
Musl's default bitlength for off_t is always 64bit therefore define off64_t as off_t on musl

Signed-off-by: Khem Raj <raj.khem@gmail.com>

### DIFF
--- a/src/malloc_hook_mmap_linux.h
+++ b/src/malloc_hook_mmap_linux.h
@@ -45,6 +45,11 @@
 #include <sys/syscall.h>
 #include <unistd.h>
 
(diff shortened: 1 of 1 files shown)


alk:
I've killed linux syscalls support. So this is not needed anymore.

alk:
Oops. I actually wrongly assumed it was about syscalls support thingy but it is not. Yet we want to be more careful here. So I need a bit of time to think here.

This is now opened as issue #1406

alk:
See #1407. Will handle there.

PR #1380, opened 2023-01-10
Remove unused using declaration

jiaweian1001:
std::sort in `heap-profiler.cc` is not in use, so remove it.


alk:
Applied. Thanks.

PR #1390, opened 2023-03-21
Fix compilcation at clang for Android x86_64

Un1oR:
https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2017/n4713.pdf#section.5.10

> In addition, some identifiers are reserved for use by C++ implementations and **shall not be used otherwise**; no
diagnostic is required.
> Each identifier that **contains a double underscore __** or begins with an underscore followed by an uppercase letter **is reserved to the implementation for any use**.

Practically at clang for Android x86_64 causes
```
error: expected member name or ';' after declaration specifiers
  int64_t            __unused[3];
  ~~~~~~~                    ^
```


alk:
I've finally dropped linux syscalls thingy. So hopefully isn't needed anymore.

PR #1391, opened 2023-03-29
Upgrade solution from Visual Studio 2015 to 2022.

Quipyowert2:
Upgraded from v140 (Visual Studio 2015) to v143 (VS2022) and SDK version 10.0.19041. 

Visual Studio versions older than 2019 are [no longer supported by Microsoft](https://devblogs.microsoft.com/visualstudio/support-ends-for-older-versions-of-visual-studio-feb2022/).


alk:
Thanks for the patch but I am not sure this is entirely right. MS said it ended mainstream support for vs2015, but "extended" support is still there. See here what this means: https://learn.microsoft.com/en-us/visualstudio/productinfo/vs-servicing

Given that it is actually straightforward to change "toolchain version", I think we should keep code/msbuild files vs2015 for now.

Quipyowert2:
I just re-read that article I linked and realized you're correct; VS2015 is still supported for a couple more years. I guess I'll close this PR.

Issue #1392, opened 2023-04-14
tcmalloc_minimal_static: CMake build failed


DuKeM-CSGO:
When running `cmake --build . --config Release -v -- -r` I get the following error.
```
  tcmalloc_minimal_static.vcxproj -> F:\Github\DKM\RoadtoCmake\build\Release\_deps\gperftools-build\Release\tcmalloc_minimal_static.lib
PostBuildEvent:
  setlocal
  F:\TDM-GCC-64\bin\objcopy.exe -W malloc -W free -W realloc -W calloc -W cfree -W memalign -W posix_memalign -W valloc -W pvalloc -W aligned_alloc -W malloc_stats -W mallopt -W mallinfo -W nallocx -W _Znwm -W _ZnwmRKSt9nothrow_t -W _Znam -W _ZnamRKSt9nothrow_t -W _ZdlPv -W _ZdaPv -W __Znwm -W __ZnwmRKSt9nothrow_t -W __Znam -W __ZnamRKSt9nothrow_t -W __ZdlPv -W __ZdaPv F:/Github/DKM/RoadtoCmake/build/Release/_deps/gperftools-build/Release/tcmalloc_minimal_static.lib
  if %errorlevel% neq 0 goto :cmEnd
  :cmEnd
  endlocal & call :cmErrorLevel %errorlevel% & goto :cmDone
  :cmErrorLevel
  exit /b %1
  :cmDone
  if %errorlevel% neq 0 goto :VCEnd
  :VCEnd
  F:\TDM-GCC-64\bin\objcopy.exe: illegal pathname found in archive member: F:/Github/DKM/RoadtoCmake/build/Release/_deps/gperftools-build/windows_object.dir/Release/preamble_patcher_with_stub.obj
H:\Program Files\Microsoft Visual Studio\MSBuild\Microsoft\VC\v170\Microsoft.CppCommon.targets(159,5): error MSB3073: Command "setlocal [F:\Github\DKM\RoadtoCmake\build\Release\_deps\gperftools-build\tcmalloc_minimal_static.vcxproj]
[...]

PR #1396, opened 2023-05-20
PRs for pytorch

xuhancn:
I'm working on improve pytorch Windows version performance, detailed info here: https://github.com/pytorch/pytorch/issues/62387

Picture1 3017

Tc_malloc library is the best performance candidate for final solution. But I need upstream two place modification.
1. gperftools has a sub-library "logging", which is same name as pytorch's another sub module , google's "[XNNPACK](https://github.com/google/XNNPACK)".
Two same name libraries in the same CMake project will cause build fail.
image
I add a "gpf_" prefix for logging library to indicate this library is belong to gperftools.

2. I will call tc_malloc and tc_free manually. And I didn't need tc_malloc automatic hook system malloc/free functions. It would break software stack balancing.

Please review and comment for this PR. If this PR was merged, I will integrate gperftools to pytorch to improve pytorch performance.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -568,7 +568,7 @@ set(LOGGING_INCLUDES
 set(liblogging_la_SOURCES src/base/logging.cc
         src/base/dynamic_annotations.c
         ${LOGGING_INCLUDES})
-add_library(logging STATIC ${liblogging_la_SOURCES})
+add_library(gpt_logging STATIC ${liblogging_la_SOURCES})
 
 set(SYSINFO_INCLUDES
         src/base/sysinfo.h
@@ -630,7 +630,7 @@ if(MINGW OR MSVC)
           src/base/atomicops-internals-x86.cc
           ${SPINLOCK_INCLUDES})
   add_library(spinlock STATIC ${libspinlock_la_SOURCES})
-  set(LIBSPINLOCK windows spinlock sysinfo logging)
+  set(LIBSPINLOCK windows spinlock sysinfo gpt_logging)
   # We also need to tell mingw that sysinfo.cc needs shlwapi.lib.
   # (We do this via a #pragma for msvc, but need to do it here for mingw).
   target_link_libraries(sysinfo shlwapi)
@@ -653,7 +653,7 @@ else()
           ${SPINLOCK_INCLUDES})
   add_library(spinlock STATIC ${libspinlock_la_SOURCES})
   target_link_libraries(spinlock ${nanosleep_libs})
-  set(LIBSPINLOCK spinlock sysinfo logging)
+  set(LIBSPINLOCK spinlock sysinfo gpt_logging)
   set(TCMALLOC_CC "src/tcmalloc.cc")
   set(SYSTEM_ALLOC_CC "src/system-alloc.cc")
 
@@ -683,7 +683,7 @@ if(BUILD_TESTING)
   # By default, MallocHook takes stack traces for use by the heap-checker.
   # We don't need that functionality here, so we turn it off to reduce deps.
   target_compile_definitions(low_level_alloc_unittest PRIVATE NO_TCMALLOC_SAMPLES)
-  target_link_libraries(low_level_alloc_unittest spinlock sysinfo logging ${maybe_threads_lib})
+  target_link_libraries(low_level_alloc_unittest spinlock sysinfo gpt_logging ${maybe_threads_lib})
   add_test(low_level_alloc_unittest low_level_alloc_unittest)
 
   set(ATOMICOPS_UNITTEST_INCLUDES src/base/atomicops.h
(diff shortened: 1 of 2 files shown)


xuhancn:
@alk could you please help on review this PR?

xuhancn:
> Also I am genuinely curious. Somehow malloc performance also affects some TF benchmarks (or used to affect few years back).
> 
> But I am quite puzzled how come ML workload which should be super-bottlenecked on number crunching stuff, actually ends up at least partially depending on memory allocator performance. Does look like possible opportunity to beef up your project's performance if you find how/if it can depend less on dynamic memory allocation.

In order to figure out the malloc performance gap, I wrote a project [bench_malloc](https://github.com/xuhancn/bench_malloc).  Please the result pic: 
image
We can find that, the most gap is in the "malloc & access" data. It is funny:
1. Malloc function acturally not real malloc memory, it is only update the memory meta data.
2. When the memory acturally malloced? When the memory was accessed. 
3. Once a memory was malloced and acturally not exist. OS will trigger page_fault exception, and than prapare the memory page for access.
4. DL tensor is a large memory area, and will trigger a lots of paga_fault.
5. If DL tensors was create/destory frequently, it would cause a big memory overheart.
6. tc_malloc/mimalloc seems have mechanism to re-use malloced memory and improve performance.

xuhancn:
> Ah, I missed part of your description above. So the logging thing appears to be some artifact of how you integrate cmake thingy. I wonder if there is cleaner way. Sure, that name thing is just name, but there should be some more principled way that doesn't have gperftools' cmake details 'leak' into whatever you're doing in your project.
> 
> As for the second commit, can you please elaborate on that "balancing" thingy? I am not familiar with this notion of "software stack balancing".

The second commit, "[module_enter_exit_hook](https://github.com/gperftools/gperftools/blob/master/src/tcmalloc.cc#L1167)" will initial and hook all loaded modules "malloc/free" & "new/delete" functions.
In some Windows scenario, it was equal "/MT" build prarmeter and embed heap memory into binary. Please reference here: https://stackoverflow.com/questions/31657775/exit-application-while-stdin-blocking-on-windows

Mimalloc has a option to enable hook functions: https://github.com/microsoft/mimalloc/blob/master/CMakeLists.txt#L10C8-L10C29 and it is default turned off.

My second commit is want to add a switch to turn hook off.

alk:
For the "hook" thingy, I think I understood. Why then not simply use WIN32_OVERRIDE_ALLOCATORS ?

For the perf thingy couple things. First, your microbenchmark might be (and likely is) not representative of actual performance of whatever workload you're optimizing. So be careful interpreting the numbers there.

Second, I don't doubt that gperftools will be in some cases faster. And yes we do occasionally behave less aggressive than others w.r.t. returning large allocations back to kernel (and then causing page faults and pages zeroings when memory is allocated back). But my point is, whatever is being re-initialized which includes those larger allocations too, shouldn't be happening as often as it appears to (and we this is the case because there is perf impact). If you apply your work towards avoiding this, you are likely to reap benefits much much bigger than playing with malloc tweakings. I.e. whatever matrices and what not should be allocated more or less once, then malloc perf and page faults etc won't matter.

xuhancn:
> For the "hook" thingy, I think I understood. Why then not simply use WIN32_OVERRIDE_ALLOCATORS ?
> 
> For the perf thingy couple things. First, your microbenchmark might be (and likely is) not representative of actual performance of whatever workload you're optimizing. So be careful interpreting the numbers there.
> 
> Second, I don't doubt that gperftools will be in some cases faster. And yes we do occasionally behave less aggressive than others w.r.t. returning large allocations back to kernel (and then causing page faults and pages zeroings when memory is allocated back). But my point is, whatever is being re-initialized which includes those larger allocations too, shouldn't be happening as often as it appears to (and we this is the case because there is perf impact). If you apply your work towards avoiding this, you are likely to reap benefits much much bigger than playing with malloc tweakings. I.e. whatever matrices and what not should be allocated more or less once, then malloc perf and page faults etc won't matter.

1. I tried to use "WIN32_OVERRIDE_ALLOCATORS", but is was occured same build fail issue.
2. tc_malloc is the best performance one is not from my bench_malloc tool, it is from existing case: https://github.com/pytorch/pytorch/issues/62387, and brief RFC here: https://github.com/pytorch/pytorch/issues/102534.
3. I have add mimalloc to pytorch now, and [PR ](https://github.com/pytorch/pytorch/pull/102595)was merged. I want tp add tc_malloc also. And then run a serise pytorch benchmark to select better one.
4. For malloc frequently, pytorch also help its submodule prepare memory, such as oneDNN. In acturaly scenario, they have a lot of reorder operations, make memory contiguous operations. These operations need a lot of temp buffer, and it is necessary.

alk:
BTW over here: https://github.com/gperftools/gperftools/issues/1392 is see what looks like more logical way to integrate gperftools with whatever "outer" cmake project. I.e. without having to rename any internal libraries etc.

You'd still be facing whatever overring issue there is, which I am unforunately not able to understand yet. But at least one of your patches won't be needed.

PR #1399, opened 2023-06-21
Fix uc_mcontext for apple silicon

### LINKED ISSUE #1398: CMake file `cmake/PCFromUContext.cmake`  is missing entry for apple silicon
Hi,

I tried to build and use the cmake version on a Mac M1 and one of the errors I encountered was a missing definition in `cmake/PCFromUContext.cmake`.

The autogen/config.sh workflow works fine, because [pc_from_ucontext.m4](https://github.com/gperftools/gperftools/blob/bf8b714bf5075d0a6f2f28504b43095e2b1e11c5/m4/pc_from_ucontext.m4#L48) defines the right override.

This issue can be fixed by adding the following line to `cmake/PCFromUContext.cmake`
```
        "uc_mcontext->__ss.__srr0"# OS X (x86, >=10.5 [untested])
        "uc_mcontext->__ss.__pc")  # OS X (arm64, >=10.5 [untested])
```

Thanks,
Henrik

henrixapp:
Closes #1398


alk:
Applied. Thanks.

PR #1427, opened 2023-09-04
fix: cmake config.h defines declaration

vaind:
This fixes how precompiler macros are defined in the cmake build. 
I've manually searched for each variable and aligned the declaration in `config.h` with how the define is actually used (be it `#if` or `#ifdef`).
In a couple of cases where the usage was inconsistent between sourcefiles (`#if` in some files, `#ifdef` in others), I've consolidated these source files to use the same logic.

Ideally, all source files would use just one way to check these, but doing such a change would cause a very large PR changelist so I didn't want to do it here just to land the fix.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -150,3 +150,4 @@
 /unique_path_unittest.exe
 /unwind_bench
 /unwind_bench.exe
+/build
--- a/cmake/config.h.in
+++ b/cmake/config.h.in
@@ -28,10 +28,10 @@
 #cmakedefine ENABLE_SIZED_DELETE
 
 /* Define to 1 if you have the <asm/ptrace.h> header file. */
-#cmakedefine01 HAVE_ASM_PTRACE_H
+#cmakedefine HAVE_ASM_PTRACE_H
 
 /* Define to 1 if you have the <cygwin/signal.h> header file. */
-#cmakedefine01 HAVE_CYGWIN_SIGNAL_H
+#cmakedefine HAVE_CYGWIN_SIGNAL_H
 
 /* Define to 1 if you have the declaration of `backtrace', and to 0 if you
    don't. */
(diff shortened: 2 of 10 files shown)


alk:
Applied. Thanks. Can you please elaborate what is the "fix" aspect of this (i.e. other than matter of taste).

vaind:
> Applied. Thanks. Can you please elaborate what is the "fix" aspect of this (i.e. other than matter of taste).

Not a matter of taste at all. It didn't compile before for "features" that were not available. Let me give you an example I've encountered when compiling for Android:

1. First, this is executed to find out whether `__environ` is defined:
https://github.com/gperftools/gperftools/blob/7ad1dc7693b7b04199681df474122248634a9808/CMakeLists.txt#L327-L334
This results in `HAVE___ENVIRON == false` because it isn't on NDK.
2. Next, the original cod had `#cmakedefine01 HAVE___ENVIRON` in `config.h.in` which will result in (see [cmake docs](https://cmake.org/cmake/help/latest/command/configure_file.html) for details): `#define HAVE___ENVIRON 0`
3. Now, the code that conditionally uses it has an `#ifdef` - https://github.com/gperftools/gperftools/blob/7ad1dc7693b7b04199681df474122248634a9808/src/base/sysinfo.cc#L160-L168
4. Thus, because `HAVE___ENVIRON` is defined (albeit to a 0), ifdef succeeds and the compilation fails because `__environ` is actually unknown.

The correct way to do this is either:
- use `#ifdef` and `#if defined()` in code but then combine that with `config.h.in` using `#cmakedefine` which will conditionally evaluate to ``#define HAVE___ENVIRON 1` if and only if the cmake variable is true
- use `#if` directly and combine it with  `#cmakedefine01` which will always make the macro defined.

Which way to use is indeed a matter of taste. Currently, this repository does both approaches which leads to confusion (and errors in this case). This PR just fixed the errors, didn't try to make the decision and unify which way to go.

alk:
Thanks for the update. And for the fix.

Issue #1318, opened 2021-12-24
build for Android/x86_64 failed


leleliu008:
https://github.com/gperftools/gperftools/releases/download/gperftools-2.9.1/gperftools-2.9.1.tar.gz
https://github.com/gperftools/gperftools.git

![7](https://user-images.githubusercontent.com/3040852/147342848-a3abf9de-0999-49a3-8709-1673b8577147.PNG)

https://github.com/gperftools/gperftools/blob/master/src/base/linux_syscall_support.h#L431

lennoxho:
Can you provide the version information of the compiler you are using?
`clang++ -v` or `g++ -v`

leleliu008:
![Screenshot from 2022-04-23 03-03-30](https://user-images.githubusercontent.com/3040852/164778070-e06c8766-cc80-4271-a755-93cd2e141101.png)

```
Android (7714059, based on r416183c1) clang version 12.0.8 (https://android.googlesource.com/toolchain/llvm-project c935d99d7cf2016289302412d708641d52d2f7ee)
Target: x86_64-unknown-linux-gnu
Thread model: posix
InstalledDir: /home/leleliu008/.ndk-pkg/android-ndk-r23b/toolchains/llvm/prebuilt/linux-x86_64/bin
Found candidate GCC installation: /home/leleliu008/.ndk-pkg/android-ndk-r23b/toolchains/llvm/prebuilt/linux-x86_64/bin/../lib/gcc/i686-linux-android/4.9.x
Found candidate GCC installation: /home/leleliu008/.ndk-pkg/android-ndk-r23b/toolchains/llvm/prebuilt/linux-x86_64/bin/../lib/gcc/x86_64-linux-android/4.9.x
Found candidate GCC installation: /usr/lib/gcc/x86_64-linux-gnu/11
Selected GCC installation: /usr/lib/gcc/x86_64-linux-gnu/11
Candidate multilib: .;@m64
Candidate multilib: 32;@m32
Candidate multilib: x32;@mx32
Selected multilib: .;@m64
```

lennoxho:
I'm not familiar with the Android platform, but this looks similar to this issue
https://stackoverflow.com/questions/9247151/compiling-icu-for-android-uint64-t-does-not-name-a-type

leleliu008:
Thanks.

alk:
linux syscall stuff is now gone. I am able to compile for with android ndk. It is not a priority for me, so not yet figured how to test.

[...]

PR #1431, opened 2023-09-11
fix: a few cmake issues

vaind:
- fix: cmake tcmalloc version
- fix: cmake test issues

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1103,7 +1103,7 @@ if(GPERFTOOLS_BUILD_HEAP_CHECKER OR GPERFTOOLS_BUILD_HEAP_PROFILER)
     add_executable(tcmalloc_unittest ${tcmalloc_unittest_SOURCES})
     target_link_libraries(tcmalloc_unittest tcmalloc ${tcmalloc_unittest_LIBADD})
     add_test(NAME tcmalloc_unittest
-            COMMAND "${CMAKE_CURRENT_SOURCE_DIR}/src/tests/tcmalloc_unittest.sh")
+            COMMAND "${CMAKE_CURRENT_SOURCE_DIR}/src/tests/tcmalloc_unittest.sh" ./tcmalloc_unittest)
 
     # This makes sure it's safe to link in both tcmalloc and
     # tcmalloc_minimal.  (One would never do this on purpose, but perhaps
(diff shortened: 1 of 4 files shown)


alk:
Hi. And thanks for taking care of cmake side of things. Do I understand correctly that this is work in progress?

vaind:
> Hi. And thanks for taking care of cmake side of things. Do I understand correctly that this is work in progress?

hey. This fixes a couple of small issues so it can already be merged. There are still some cmake tests that fail (with segfaults) but I don't have fixes for those because they require a deeper look into what's wrong. If I ever do have a look and fix some, I'll post those in a separate PR. Weirdly, same tests work fine when built via autoconf/make on the same machine (WSL2).

alk:
Thanks for update. So on first look first commit seems okay (but I need a closer look). Second, with some arbitrary additions like "fflush' and so on, sorry I cannot take as-is.

As for failing tests, my quick look-based understanding is this is messed up visibility and linking details "stuff". Tests seem to link libraries (with internals hidden) and then same internals linked in test-only which effs-up all kinds of things. At least on quick glance it was my impression.

And our tests and ABI is indeed a bit of trouble and I have some grand plans there. So my plan is to first sort that and look at cmake later (if at all; another option is to drop it, since I haven't seen much interest there).

vaind:
> with some arbitrary additions like "fflush' and so on, sorry I cannot take as-is.

I'm open to suggestions. This is there because if a test fails, you wouldn't actually see the most recently printed output because it's just aborted. I got hit by this here, specifically: https://github.com/gperftools/gperftools/pull/1431/files#diff-abe99e3fd6e08467c3f6f5f6868ddf38f29727d3e59f503295fe38c79c82c1fbR1599

> So my plan is to first sort that and look at cmake later (if at all; another option is to drop it, since I haven't seen much interest there).

Alright, I understood CMake is not really a priority in this project, although I am quite an advocate as it does help with cross-platform support a bit.
Honestly, I'd be fine with using autoconf & make, it's just that I'm trying to make this work for Android (see [also my comment in another issue](https://github.com/gperftools/gperftools/issues/1318#issuecomment-1714052459)) and CMake is kinda something that's available out of the box in gradle and the android plugin.

alk:
Hm. So it doesn't seem to work. Here is what I get when I update CMakeFile to 2.12.90

```
/* Define the version number so folks can check against it */
#define TC_VERSION_MAJOR  2
#define TC_VERSION_MINOR  12
#define PROJECT_VERSION_PATCH
#ifdef PROJECT_VERSION_PATCH
  #define TC_VERSION_PATCH ".x" PROJECT_VERSION_PATCH
#else
  #define TC_VERSION_PATCH ""
#endif
#define TC_VERSION_STRING "gperftools 2.12.90"
```

vaind:
> Hm. So it doesn't seem to work. Here is what I get when I update CMakeFile to 2.12.90
> 
> ```
> /* Define the version number so folks can check against it */
> #define TC_VERSION_MAJOR  2
> #define TC_VERSION_MINOR  12
> #define PROJECT_VERSION_PATCH
> #ifdef PROJECT_VERSION_PATCH
>   #define TC_VERSION_PATCH ".x" PROJECT_VERSION_PATCH
> #else
>   #define TC_VERSION_PATCH ""
> #endif
> #define TC_VERSION_STRING "gperftools 2.12.90"
> ```

That looks like an old version of this PR with some temp code I've accidentally committed.

leleliu008:
@vaind 

You could use perl via https://termux.dev 

If you only want to use `adb shell`, you could download my prebuild perl for android (https://github.com/leleliu008/ndk-pkg-package-manually-build/actions/runs/6155655219)

```bash
tar vxf perl-5.34.0-android28-aarch64.tar.xz
tar cvf perl-5.34.0-android28-aarch64.tar.gz perl-5.34.0-android28-aarch64 --gzip
adb push perl-5.34.0-android28-aarch64.tar.gz /data/local/tmp
```

```bash
export LD_LIBRARY_PATH=/data/local/tmp/perl-5.34.0-android28-aarch64/lib/perl5/5.34.0/aarch64-android/CORE
```

alk:
> That looks like an old version of this PR with some temp code I've accidentally committed.

This was exactly first commit in this pull request.

vaind:
> > That looks like an old version of this PR with some temp code I've accidentally committed.
> 
> This was exactly first commit in this pull request.

exactly. individual commits in this PR may not be cherry-pickable (the first one isn't). there's another commit that fixes the issue above.

Unless explicitly stated otherwise in the PR description (and only where it looks useful), all PRs I create are meant to be taken as a whole (& preferably squashed). 

But if you really wanted to cherry-pick from this, you'd need the first and the third commit and squash them

alk:
Hi. Gentle ping on previous comment.

vaind:
PR abandoned/stale. Feel free to use any of the code in case you want to land this.

Issue #1332, opened 2022-04-19
Implicit TLS detection in CMakeLists.txt does not account for MSVC


lennoxho:
`__thread` is not supported on MSVC. 
We should add an `elseif(MSVC)` clause that checks for `__declspec(thread)` instead.

https://github.com/gperftools/gperftools/blob/fe85bbdf4cb891a67a8e2109c1c22a33aa958c7e/CMakeLists.txt#L357-L367

alk:
Sure. Go ahead and add, please.

lennoxho:
https://github.com/gperftools/gperftools/pull/1435

lennoxho:
Merged https://github.com/gperftools/gperftools/pull/1439

Issue #1434, opened 2023-09-17
Also expose SetMemoryReleaseRate and GetMemoryReleaseRate as C shims


lennoxho:

lennoxho:
https://github.com/gperftools/gperftools/pull/1435

lennoxho:
https://github.com/gperftools/gperftools/pull/1439

Issue #1433, opened 2023-09-17
Fix page_heap_test and mmap_hook_test in the Windows CMake build


lennoxho:
As seen in https://github.com/gperftools/gperftools/issues/1189, a few adjustments is needed to make the Windows build work.

lennoxho:
https://github.com/gperftools/gperftools/pull/1435

lennoxho:
New PR https://github.com/gperftools/gperftools/pull/1440

PR #1435, opened 2023-09-17
Couple of fixes to Windows CMake build; Add couple of missing C-shims for malloc extension API

lennoxho:
This PR includes 3 fixes/improvements
1. https://github.com/gperftools/gperftools/issues/1433 by porting over the VSProj changes https://github.com/gperftools/gperftools/commit/e521472f1ac1ecf028e8c922ca4f8741473afd66
    - ALL_BUILD can now be completed successfully
3. Fix https://github.com/gperftools/gperftools/issues/1332
5. https://github.com/gperftools/gperftools/issues/1434

Tested with gcc 12.3 on Linux (should be a no-opt) and MSVC 19.31 on Windows.
With (2)

Before:
```
-- Performing Test HAVE_TLS
-- Performing Test HAVE_TLS - Failed
```
After:
```
-- Performing Test HAVE_TLS
-- Performing Test HAVE_TLS - Success
```


alk:
Should we just hardcode HAVE_TLS for MSVC? Is there any windows platforms that lack it? Likely to be simpler no?

For the rate thingy, can you please consider going with MallocExtension_{Get,Set}NumericProperty. Reason I ask is our ABI is already a bit of a mess. I am super-eager to expose one more implementation detail in a "hard" api like this. Numeric property instead is "soft" (that is, if some future version ignores client's request for release rate, at least client's app won't crash or fail to handle dynamic linking).

Please, let me know what you think.

lennoxho:
Hello!

> Should we just hardcode HAVE_TLS for MSVC? Is there any windows platforms that lack it? Likely to be simpler no?

Hmm, I mean it's highly unlikely that we'll encounter a toolchain/platform that lacks implicit TLS (although I couldn't find an authoritative source on this).

On the other hand, the new check for MSVC is fairly trivial and consistent with what we are doing for gcc...
So I'm tempted to slightly err on the side of caution here.

> For the rate thingy, can you please consider going with MallocExtension_{Get,Set}NumericProperty. Reason I ask is our ABI is already a bit of a mess. I am not super-eager to expose one more implementation detail in a "hard" api like this. Numeric property instead is "soft" (that is, if some future version ignores client's request for release rate, at least client's app won't crash or fail to handle dynamic linking).

That makes sense. However, `MallocExtension_{Get,Set}NumericProperty`'s API only works with `size_t`s. 
The `MemoryReleaseRate` property deals with `double`s. Or is there another interface I'm missing?

With that said, I'm fine with just dropping the C-shim addition. It's mostly prompted by laziness :P - we just couldn't be bothered to `dlsym` the mangled name.

lennoxho:
Hello!

> Should we just hardcode HAVE_TLS for MSVC? Is there any windows platforms that lack it? Likely to be simpler no?

Hmm, I mean it's highly unlikely that we'll encounter a toolchain/platform that lacks implicit TLS (although I couldn't find an authoritative source on this).

On the other hand, the new check for MSVC is fairly trivial and consistent with what we are doing for gcc...
So I'm tempted to slightly err on the side of caution here.

> For the rate thingy, can you please consider going with MallocExtension_{Get,Set}NumericProperty. Reason I ask is our ABI is already a bit of a mess. I am not super-eager to expose one more implementation detail in a "hard" api like this. Numeric property instead is "soft" (that is, if some future version ignores client's request for release rate, at least client's app won't crash or fail to handle dynamic linking).

That makes sense. However, `MallocExtension_{Get,Set}NumericProperty`'s API only works with `size_t`s. 
The `MemoryReleaseRate` property deals with `double`s. Or is there another interface I'm missing?

With that said, I'm fine with just dropping the C-shim addition. It's mostly prompted by laziness :P - we just couldn't be bothered to `dlsym` the mangled name.

lennoxho:
Merged into PR https://github.com/gperftools/gperftools/pull/1439 as requested.

Issue #1436, opened 2023-09-17
Add the ability to disable TCMalloc replacement on Windows via environment variable


lennoxho:
The TBBMalloc library provides a means to disable replacement of `malloc` et al. during program startup via an [environment variable](https://oneapi-src.github.io/oneTBB/main/tbb_userguide/Windows_C_Dynamic_Memory_Interface_Replacement.html)

This feature has proven to be very useful for debugging/unblocking issues related to runtime replacement of allocation functions.
Here's a recent example of such an issue https://github.com/oneapi-src/oneTBB/issues/665.
(I'm surprised nobody has opened a case here yet on that particular issue. It most certainly affects TCMalloc too).

lennoxho:
https://github.com/gperftools/gperftools/pull/1437

lennoxho:
https://github.com/gperftools/gperftools/pull/1439

PR #1437, opened 2023-09-17
Add the ability to disable TCMalloc replacement on Windows via environment variable

lennoxho:
This PR [adds the ability to disable TCMalloc replacement on Windows during program startup by specifying](https://github.com/gperftools/gperftools/issues/1436) `TCMALLOC_DISABLE_REPLACEMENT=1`

As discussed in the linked issue, this is a useful feature that is offered by TBBMalloc.

### DIFF
--- a/src/windows/patch_functions.cc
+++ b/src/windows/patch_functions.cc
@@ -92,6 +92,7 @@
 #include <vector>
 #include <base/logging.h>
 #include "base/spinlock.h"
+#include "getenv_safe.h"  // for TCMallocGetenvSafe
 #include "gperftools/malloc_hook.h"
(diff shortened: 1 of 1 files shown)


alk:
Thanks for the patch. In principle I dont mind something like this. But I am slightly concerned that this feature will forever be windows-patching-only.

I am not even sure we will keep supporting patching thingy going forward (well, MS's mimalloc supporting patching is probably good sign that we can, but whether we have human capacity to maintain it will remain to be a question, potentially)

Give me couple days to ponder.

lennoxho:
Merged into https://github.com/gperftools/gperftools/pull/1439 as requested.

PR #1438, opened 2023-09-17
Provide a more performant SpinLockDelay implementation on Windows 8.0 on newer

lennoxho:
Issue: https://github.com/gperftools/gperftools/issues/1333

This PR provides a more performant SpinLockDelay implementation on Windows 8.0 or newer based on [`WaitOnAddress`](https://learn.microsoft.com/en-us/windows/win32/api/synchapi/nf-synchapi-waitonaddress) and friends.

`WaitOnAddress` is roughly equivalent to Linux's futex, which is what TCMalloc is currently using on that platform.

There was a previous PR attempt https://github.com/gperftools/gperftools/pull/1350.

This PR addresses a comment to maintain support for Windows 7 or older.
That comment is addressed here by checking `_WIN32_WINNT` during build time.

The TCMalloc user can opt out of this change by using an older version of the Windows SDK or by specifying [`_WIN32_WINNT`](https://learn.microsoft.com/en-us/cpp/porting/modifying-winver-and-win32-winnt?view=msvc-170) to an older value.

### DIFF
--- a/src/base/spinlock_win32-inl.h
+++ b/src/base/spinlock_win32-inl.h
@@ -40,6 +40,7 @@ namespace internal {
 
 void SpinLockDelay(std::atomic<int> *w, int32 value, int loop) {
   (void)w;
+  (void)value;
 
 // 0x0602 corresponds to Windows 8.0
 #if defined(_WIN32_WINNT) && _WIN32_WINNT >= 0x0602
--- a/src/base/spinlock_win32-inl.h
+++ b/src/base/spinlock_win32-inl.h
@@ -39,15 +39,37 @@ namespace base {
 namespace internal {
 
 void SpinLockDelay(std::atomic<int> *w, int32 value, int loop) {
+  (void)w;
+
+// 0x0602 corresponds to Windows 8.0
+#if defined(_WIN32_WINNT) && _WIN32_WINNT >= 0x0602
+  if (loop != 0) {
+    auto wait_ns = static_cast<uint64_t>(base::internal::SuggestedDelayNS(loop)) * 16ull;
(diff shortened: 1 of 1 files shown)


alk:
This is great. Thanks!

Couple things. Can you please squash commits into 1 single commit?

That stuff about auto wait_{ms,ns} assigned result of uint64_ times "ull" constant is sufficiently head-scratching IMHO. Can you please make types explicit. And perhaps drop ull?

Also that you're having to duplicate very odd/arbitrary logic from linux futex codes is worth a think. No big deal, but if you have time consider refactoring this more to unify the logic (if not, fine, I'll handle myself eventually).

Also consider checking what abseil is doing (they might use similar patch or maybe you can steal some ideas from there).

Thanks again!

P.S. Since last time waitonaddress was proposed pretty much everyone dropped support for windows 7. I just checked. But I think keeping ifdef might be needed if msvc defaults continue to be older windows SDKs (I have no idea). But just in case. We could and perhaps should consider just doing windows 8 or even windows 10 or later if that simplifies anything.

alk:
BTW, ms's own c++ library does the both compile time ifdef thingy and runtime thingy: https://github.com/microsoft/STL/blob/8f67ece4a654625220642c92fa029a5cfc77afa8/stl/src/atomic_wait.cpp#L227 (for c++20 atomic futex-like ops).

Just pointing this out. Please don't consider adding anywhere as much complexity as over there. If we can somehow simplify and hard-depend on newer API, I recommend doing that.

lennoxho:
Merged into PR https://github.com/gperftools/gperftools/pull/1439 as requested.

PR #1439, opened 2023-09-17
Fixes and improvement to Windows CMake build; Add ability to disable TCMalloc on Windows during startup; Improve spinlock performance on Windows

lennoxho:
Merging 3 PRs into one as requested : https://github.com/gperftools/gperftools/pull/1435 https://github.com/gperftools/gperftools/pull/1437 https://github.com/gperftools/gperftools/pull/1438

Issues addressed:
1. https://github.com/gperftools/gperftools/issues/1433 by porting over the VSProj changes https://github.com/gperftools/gperftools/commit/e521472f1ac1ecf028e8c922ca4f8741473afd66
2. [MERGED] Fix https://github.com/gperftools/gperftools/issues/1332
3. [MERGED] https://github.com/gperftools/gperftools/issues/1434
4. [MERGED] https://github.com/gperftools/gperftools/issues/1436
5. Provide a more performant `SpinLockDelay` implementation on Windows or newer to address https://github.com/gperftools/gperftools/issues/1333

Tested with gcc 12.3 on Linux (should be a no-opt) and MSVC 19.31 on Windows.

------------

**More details:**

1. ALL_BUILD can now be completed successfully

2. 
Before:
```
-- Performing Test HAVE_TLS - Failed
```
After;
```
-- Performing Test HAVE_TLS - Success
```

Tested with 
- gcc 12.3 on Linux (Ubuntu 22.04.1 LTS)
- MSVC 19.31 (from EWDK 22621_230511-2102) on Windows 10 22H2.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -567,6 +567,7 @@ if(MINGW OR MSVC)
   # (We do this via a #pragma for msvc, but need to do it here for mingw).
   target_link_libraries(sysinfo shlwapi)
   # spinlock uses WaitOnAddress et al. We need to link to synchronization.lib
+  # (We also do this via a #pragma for msvc, but need to do it here for mingw).
   target_link_libraries(spinlock synchronization)
 
 else()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -854,6 +854,7 @@ if(BUILD_TESTING)
           src/tests/mmap_hook_test.cc
           src/mmap_hook.cc)
   if(MSVC)
+    # Work around unresolved symbol from src/windows by linking against the entire library
     target_link_libraries(mmap_hook_test tcmalloc_minimal Threads::Threads)
   else()
     target_link_libraries(mmap_hook_test spinlock sysinfo logging)
@@ -892,7 +893,9 @@ if(BUILD_TESTING)
 
   add_executable(page_heap_test src/tests/page_heap_test.cc)
   if(MSVC)
-    # Temporary hack as discussed in commit e521472
+    # page_heap_test was changed such that it now refers to an unexported symbol.
+    # Temporary workaround by linking to the .obj files instead of tcmalloc_minimal.
+    # Also see commit e521472 for the VSProj changes.
     target_link_libraries(page_heap_test PRIVATE tcmalloc_minimal_internal Threads::Threads)
   else()
     target_link_libraries(page_heap_test tcmalloc_minimal)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -853,7 +853,11 @@ if(BUILD_TESTING)
   add_executable(mmap_hook_test
           src/tests/mmap_hook_test.cc
           src/mmap_hook.cc)
-  target_link_libraries(mmap_hook_test spinlock sysinfo logging)
+  if(MSVC)
+    target_link_libraries(mmap_hook_test tcmalloc_minimal Threads::Threads)
+  else()
+    target_link_libraries(mmap_hook_test spinlock sysinfo logging)
+  endif()
   add_test(mmap_hook_test mmap_hook_test)
 
   set(malloc_extension_test_SOURCES src/tests/malloc_extension_test.cc
@@ -888,7 +892,8 @@ if(BUILD_TESTING)
 
   add_executable(page_heap_test src/tests/page_heap_test.cc)
   if(MSVC)
-    target_link_libraries(page_heap_test tcmalloc_minimal_static)
+    # Temporary hack as discussed in commit e521472
+    target_link_libraries(page_heap_test PRIVATE tcmalloc_minimal_internal Threads::Threads)
   else()
     target_link_libraries(page_heap_test tcmalloc_minimal)
   endif()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -566,6 +566,8 @@ if(MINGW OR MSVC)
   # We also need to tell mingw that sysinfo.cc needs shlwapi.lib.
   # (We do this via a #pragma for msvc, but need to do it here for mingw).
   target_link_libraries(sysinfo shlwapi)
+  # spinlock uses WaitOnAddress et al. We need to link to synchronization.lib
+  target_link_libraries(spinlock synchronization)
 
 else()
   set(SPINLOCK_INCLUDES src/base/spinlock.h
--- a/src/base/spinlock_win32-inl.h
+++ b/src/base/spinlock_win32-inl.h
@@ -32,7 +32,9 @@
  * This file is a Win32-specific part of spinlock_internal.cc
  */
 
+
 #include <windows.h>
+#pragma comment(lib, "Synchronization.lib")
 
 namespace base {
 namespace internal {
--- a/src/base/spinlock_win32-inl.h
+++ b/src/base/spinlock_win32-inl.h
@@ -35,19 +35,50 @@
 
 #include <windows.h>
 
+// 0x0602 corresponds to Windows 8.0
+#if defined(_WIN32_WINNT) && _WIN32_WINNT >= 0x0602
+#   define HAVE_WAIT_ON_ADDRESS
+#endif
+
+#ifdef HAVE_WAIT_ON_ADDRESS
+#   pragma comment(lib, "Synchronization.lib")
+#endif
+
 namespace base {
 namespace internal {
 
 void SpinLockDelay(std::atomic<int> *w, int32 value, int loop) {
+  (void)w;
+  (void)value;
+
+#ifdef HAVE_WAIT_ON_ADDRESS
+  if (loop != 0) {
+    auto wait_ns = static_cast<uint64_t>(base::internal::SuggestedDelayNS(loop)) * 16;
+    auto wait_ms = wait_ns / 1000000;
+
+    WaitOnAddress(w, &value, 4, static_cast<DWORD>(wait_ms));
+  }
+#else
   if (loop == 0) {
   } else if (loop == 1) {
     Sleep(0);
   } else {
     Sleep(base::internal::SuggestedDelayNS(loop) / 1000000);
   }
+#endif
 }
 
 void SpinLockWake(std::atomic<int> *w, bool all) {
+  (void)w;
+  (void)all;
+
+#ifdef HAVE_WAIT_ON_ADDRESS
+  if (all) {
+    WakeByAddressAll((void*)w);
+  } else {
+    WakeByAddressSingle((void*)w);
+  }
+#endif
 }
 
 } // namespace internal
--- a/src/base/spinlock_win32-inl.h
+++ b/src/base/spinlock_win32-inl.h
@@ -32,53 +32,26 @@
  * This file is a Win32-specific part of spinlock_internal.cc
  */
 
-
 #include <windows.h>
-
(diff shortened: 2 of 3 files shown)


lennoxho:
Addressing review comments from the merged PRs,

> Should we just hardcode HAVE_TLS for MSVC? Is there any windows platforms that lack it? Likely to be simpler no?

Hmm, I mean it's highly unlikely that we'll encounter a toolchain/platform that lacks implicit TLS (although I couldn't find an authoritative source on this).

On the other hand, the new check for MSVC is fairly trivial and consistent with what we have been doing for gcc...
So I'm tempted to slightly err on the side of caution here.

> For the rate thingy, can you please consider going with MallocExtension_{Get,Set}NumericProperty. Reason I ask is our ABI is already a bit of a mess. I am not super-eager to expose one more implementation detail in a "hard" api like this. Numeric property instead is "soft" (that is, if some future version ignores client's request for release rate, at least client's app won't crash or fail to handle dynamic linking).

That makes sense. However, MallocExtension_{Get,Set}NumericProperty's API only works with `size_t`s.
The MemoryReleaseRate property deals with doubles. Or is there another interface I'm missing?

With that said, I'm fine with just dropping the C-shim addition. It's mostly prompted by laziness :P - we just couldn't be bothered to dlsym the mangled name.

> But I am slightly concerned that this feature will forever be windows-patching-only. I am not even sure we will keep supporting patching thingy going forward (well, MS's mimalloc supporting patching is probably good sign that we can, but whether we have human capacity to maintain it will remain to be a question, potentially)

I can understand that. In fact I brought the same issue up with the TBBMalloc maintainers in the past.

A couple of points to consider
1. Beyond MSFT's minmalloc, MSFT also ships https://github.com/microsoft/Detours, which is their blessed way to perform runtime function interception
    - Slightly off topic, but this could be a more stable way to do trampolines on Windows. It would most definitely save us the hassle and stress of maintaining our own implementation.
    - The Detours library also provides the means to [DetourCreateProcessWithDlls](https://github.com/microsoft/Detours/wiki/DetourCreateProcessWithDlls). This could in theory be used to provide a post-build time means to inject alternative allocators to existing binaries, sort of like Linux's `LD_PRELOAD`
 2. As noted in the [issue](https://github.com/gperftools/gperftools/issues/1436), the ability to disable function interception after shipping has proven to be invaluable in unblocking end users when MSFT accidently broke all allocation libraries on Windows 11.
    - On our end it was a mad scramble to find a fix (we ultimately added interception of `RtlAllocateHeap` to work around the issue).
 
> That stuff about auto wait_{ms,ns} assigned result of uint64_ times "ull" constant is sufficiently head-scratching IMHO. Can you please make types explicit. And perhaps drop ull?

I've removed all `u` and `ull` suffixes. Sorry, force of habit since we enable `W4` on Windows.

> Also that you're having to duplicate very odd/arbitrary logic from linux futex codes is worth a think. No big deal, but if you have time consider refactoring this more to unify the logic (if not, fine, I'll handle myself eventually).

Umm, please do :)

> Also consider checking what abseil is doing (they might use similar patch or maybe you can steal some ideas from there).

So Abseil is still using the good old `Sleep` implementation. So maybe they could steal from gperftools for a change.

> Since last time waitonaddress was proposed pretty much everyone dropped support for windows 7. I just checked. But I think keeping ifdef might be needed if msvc defaults continue to be older windows SDKs (I have no idea). But just in case. We could and perhaps should consider just doing windows 8 or even windows 10 or later if that simplifies anything.

Hmm, that would be tempting, but it feels like we should probably consult the gpertools users before making OS decisions?

alk:
I am about to take all but "fix cmake build" commits. For that commit, I'll need a bit more details why it is needed and what it is doing (sorry, I am keeping myself cmake-ignorant mostly). Feel free to add those to commit message too.

Also, notably, I see that new waitonaddress thingy is sadly not getting used in mingw. Checking if/how I can get this ON as well.

alk:
So, mingw does have synchronization not (capital letter) Synchronization. I know windows is case-insensitive but mingw isn't. I am still figuring out how to enable it. Can you please update your patch ?

alk:
Ah, well, mingw doesn't know this pragma anyways. So perhaps not worth bothering with de-capitalization.

alk:
Okay, I'll handle mingw myself as a followup to your commit. It simply makes it available without any winnt defines. But requires explicit -lsynchronization.

alk:
okay. So I took "tls detection", "expose SetMemoryRelease" and "add ability to disable replacement" commits. Thanks a lot.

For WaitOnAddress thingy I am really confused about this WIN32_WINNT define. I tested this in MSVC (via project files we ship). I was expecting this define be somehow autodefined but this wasn't the case. And just using WaitOnAddress simply works. Should we just simply use it with no ifdefs? How our users are/were supposed to use this win32_winnt define?

lennoxho:
Hi!

> ..."fix cmake build" commits. For that commit, I'll need a bit more details why it is needed and what it is doing...

Yeah I should have added a more descriptive commit message.
I've updated the PR.

Essentially `page_heap_test` was previously changed such that it now has a dependency on a non-exported symbol 
```
page_heap_test.obj : error LNK2019: unresolved external symbol "private: void __cdecl SpinLock::SlowLock(void)" (?Slo
wLock@SpinLock@@AEAAXXZ) referenced in function "void __cdecl `anonymous namespace'::TestPageHeap_Limit(void)" (?TestPa
geHeap_Limit@?A0x93b7a5d0@@YAXXZ) [D:\lenno\source\gperftools\windows_build\page_heap_test.vcxproj]
```
This didn't cause issues on Linux since all symbols have public visibility by default on the GNU toolchain.
MSVC defaults to non-exported, hence the unresolved symbol.

The workaround as noted in the VSProj change https://github.com/gperftools/gperftools/commit/e521472 was to simply link to `tcmalloc_minimal_internal`, instead of `tcmalloc_minimal`.
This works because the former is an OBJECT library (just the collection of .obj files) while the latter is an actual library.

It's a similar story with `mmap_hook_test`,
```
mmap_hook.obj : error LNK2019: unresolved external symbol WriteToStderr referenced in function "void __cdecl tcmalloc
::UnHookMMapEvents(class tcmalloc::MappingHookSpace *)" (?UnHookMMapEvents@tcmalloc@@YAXPEAVMappingHookSpace@1@@Z) [D:\
lenno\source\gperftools\windows_build\mmap_hook_test.vcxproj]
```
This is caused by the `RAW_CHECK` assertion, which references `WriteToStderr` under `src/windows`.
`WriteToStderr` is exported, but since we only link this test to `spinlock`, `sysinfo` and `logging`, it's not available at build time.

The solution is to simply link against the entire `tcmalloc_minimal` libraries.

lennoxho:
> For WaitOnAddress thingy I am really confused about this WIN32_WINNT define. I tested this in MSVC (via project files we ship). I was expecting this define be somehow autodefined but this wasn't the case. And just using WaitOnAddress simply works. Should we just simply use it with no ifdefs? How our users are/were supposed to use this win32_winnt define?

So the usage model with the Windows SDK is users are encouraged to always grab the latest and greatest versions.
Now since Windows SDK are versioned similarly to Windows versions (eg: 10.0.14393 for Windows Server 2016) this would cause compatibility issues.

So, to tell MSVC/Windows SDK to limit itself to symbols/features that are only available on their target Windows versions, we need to [define _WIN32_WINNT to the appropriate value](https://learn.microsoft.com/en-us/cpp/porting/modifying-winver-and-win32-winnt?view=msvc-170) (`0x0A00` for Windows 10).

The reason why it works out of the box for you is because `WINVER/_WIN32_WINNT` is usually predefined by your MSVC toolchain to the host platform (and I'm going to assume it's Windows 10/Windows Server 2019 in your case).
Interestingly the Enterprise WDK environment does not define `WINVER/_WIN32_WINNT`.

The rationale for gating the new spinlock implemented behind `_WIN32_WINNT` is gperftools users who need to ship their product on oldest platforms should already explicitly define `WINVER/_WIN32_WINNT` (if they are not doing that they really should).

alk:
Thanks for the details. For the sake of simplification lets just require WaitOnAddress to be always available. Very little chance we'll test and otherwise properly maintain alternative implementation.

So for that we simply define _WIN32_WINNT to match win8 value in config.h for msvc ? (And then cmake whatevs)

holenno1:
> So for that we simply define _WIN32_WINNT to match win8 value in config.h for msvc ? (And then cmake whatevs)

So that won't be necessary, as we are not assuming the user has a recent enough toolchain/SDK that contains `WaitOnAddress`.

alk:
I _want_ as to assume new enough SDK. Otherwise there is just alternative "sleep" path that we'll not test in practice and just plain violation of principle of least surprise.

That works for you?

holenno1:
Right, ok you're saying remove the `#else` and the old code path.
That makes sense. Let me do that

lennoxho:
Alright, I've updated the PR.

Tested both CMake and VSProj builds.

Can you also make sure the mingw build works?
Thanks!

alk:
Thanks.

So autotools+mingw needs a small patch. I can handle it myself (we'll lose bisectability, but mingw isn't our most prime target; so, one "bad" commit I can afford). Or you can add yourself:

```
diff --git a/Makefile.am b/Makefile.am
index 7757c68..14aab68 100644
--- a/Makefile.am
+++ b/Makefile.am
@@ -205,6 +205,7 @@ noinst_LTLIBRARIES += libspinlock.la
 libspinlock_la_SOURCES = src/base/spinlock.cc \
                          src/base/spinlock_internal.cc \
                          $(SPINLOCK_INCLUDES)
+libspinlock_la_LIBADD = -lsynchronization
 
 LIBSPINLOCK = libwindows.la libspinlock.la libsysinfo.la liblogging.la
 
diff --git a/src/base/spinlock_win32-inl.h b/src/base/spinlock_win32-inl.h
index 00c8ec6..33935b0 100644
--- a/src/base/spinlock_win32-inl.h
+++ b/src/base/spinlock_win32-inl.h
@@ -34,7 +34,10 @@
 
 
 #include <windows.h>
+
+#ifdef _MSC_VER
 #pragma comment(lib, "Synchronization.lib")
+#endif
 
 namespace base {
 namespace internal {
```

I am not able to test cmake+mingw (or call me cmake-ignorant). Not testing this combination is no big deal, IMO.

Is it too much for me to ask you to rebase and squash "fixup" commits into originals? I.e. we'd get 2 commits, one with cmake fixes, other with WaitOnAddress. (If you don't want to do this, fine, just let me know).

Also do we need to define WIN32_WINNT somewhere? Since you mentioned some sdks don't auto-enable latest and greatest (if I understood you right).

lennoxho:
I can integrate the mingw changes. No worries (you can all me an automake-skeptic).

> Is it too much for me to ask you to rebase and squash "fixup" commits into originals? I.e. we'd get 2 commits

Let me close this PR, refork, then submit a new PR.

> Also do we need to define WIN32_WINNT somewhere? Since you mentioned some sdks don't auto-enable latest and greatest (if I understood you right).

Nope, shouldn't be necessary.

With those changes 
- Users with an old Windows SDK won't be able to build gperftools.
- Users with newer SDKs will be able to successfully complete a build without setting `_WIN32_WINNT`.
- In either case if the user sets `_WIN32_WINNT` to Windows 7 or older, then they won't be able to build gperftools.

lennoxho:
New PR https://github.com/gperftools/gperftools/pull/1440

PR #1442, opened 2023-09-20
Use main executable from command line directly

iDings:
For embedded environment, pprof is used on host to map address to symbol, and pprof detect main executable from map info use full main path, for embedded environment, this is mostly inconformity.

### DIFF
--- a/src/pprof
+++ b/src/pprof
@@ -73,6 +73,7 @@ use warnings;
 use Getopt::Long;
 use Cwd;
 use POSIX;
+use File::Basename;
 
 my $PPROF_VERSION = "2.0";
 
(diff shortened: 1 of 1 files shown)


alk:
thanks for the patch but we're not maintaining perl pprof implementation that comes with our soft. Users are strongly encouraged to used go pprof from github.com/google/pprof (we share this binary with Golang community + Google uses somewhat extended version of that same code base internally)

I am also unsure and unconvinced by motivation of this change but I guess we can save time on this conversation given what I said above.

Closing as wontfix (but feel free to reopen if you have arguments why it should be taken)

iDings:
Thanks for review,it's fine to drop this PR and  i will try github.com/google/pprof.

PR #1447, opened 2023-09-25
generic_fp stacktrace: check frame size threshold for initial frame

blahgeek:


alk:
Makes sense. LGTM. Applied. Thanks

PR #1449, opened 2023-10-11
Fix buffer overflow in generic_fp stacktrace

blahgeek:
In the 'with ucontext' case, the `skip_count` would be reset to 0, and `max_depth` should not be modified. Otherwise the result array would overflow and lead to crash (happens when the stack depth is greater than max depth, or the stack is bogus)

### DIFF
--- a/src/stacktrace_generic_fp-inl.h
+++ b/src/stacktrace_generic_fp-inl.h
@@ -123,8 +123,6 @@ int capture(void **result, int max_depth, int skip_count,
             int *sizes) {
   int i = 0;
 
-  max_depth += skip_count;
-
   if (initial_pc != nullptr) {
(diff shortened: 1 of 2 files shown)


alk:
Thanks a lot. Are able to also offer unit test coverage for the issue (would be most perfect)

blahgeek:
Added unit test coverage. It can be reproduced with `TCMALLOC_STACKTRACE_METHOD=generic_fp ./stacktrace_unittest`

Also just curious: are the unit tests automatically tested in some CI pipeline? With different stacktrace methods?

alk:
Applied. Thanks a lot!

As for your question. I am manually running tests on various platforms and with various build configurations from time to time. Yes, time for us to automagically ensure that different stacktracing methods are tested. I'll consider something sooon.

PR #1455, opened 2023-10-30
AARCH64 >= armv8.3-a supports pointer authentication. If this feature is

JulianMarcusSchroeder:
enabled it modifies the previously unused upper address bits in apointer. The affected bits need to be cleared in order for stacktrace to work.

### DIFF
--- a/src/stacktrace_generic_fp-inl.h
+++ b/src/stacktrace_generic_fp-inl.h
@@ -77,6 +77,27 @@ namespace stacktrace_generic_fp {
 #define PAD_FRAME
 #endif
 
+#if __aarch64__
+// Aarch64 has pointer authentication and uses the upper 16bit of a stack
+// or return address to sign it. These bits needs to be strip in order for
+// stacktraces to work.
+void *strip_PAC(void* _ptr) {
+  void *ret;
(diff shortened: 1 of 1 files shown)


alk:
Thanks for the patch. This is great. But why bother with asm? We're not that performance-sensitive in this code, so we can simply clear those top bits in portable C++ and be done?

alk:
Actually, asm probably makes sense. This way with _only_ strip if computer is running with pointer auth supported _and_ enabled.

But looking closer I am now confused why asm part is not  under __aarch64__. We don't want x86 compilers to see this, right ?

alk:
Also you sure we require hard-coding x30 register? Can we somehow detect if arm 8.3 is enabled and then use proper XPACI instruction and let compiler choose register.

JulianMarcusSchroeder:
The xpaclri instruction is for the LR register (x30) only. It doesn't work for other registers.
I'll  add the __aarch64__  wrapper. I spaced out on that one ;-).

alk:
Applied. Thanks. Note, I had to modify commit message to add subject line, but the commit is 100% yours otherwise.

PR #1464, opened 2023-12-07
Add mallinfo2 function

### LINKED ISSUE #1414: we should implement mallinfo2
As was pointed out in https://github.com/gperftools/gperftools/issues/227#issuecomment-890569212 there is now mallinfo2 in glibc. Which solves some issues with truncated stat fields. We already implement mallinfo and so we should also do mallinfo2 when available.

m-fila:
Implemented the function `mallinfo2` that is a fixed version of `mallinfo`.   
The `mallinfo2` is part of glibc since 2.33 and orginal `mallinfo` is now deprecated.

Closes #1414

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -168,6 +168,7 @@ check_c_source_compiles("#include <stdlib.h>
 
 set(CMAKE_EXTRA_INCLUDE_FILES "malloc.h")
 check_type_size("struct mallinfo" STRUCT_MALLINFO LANGUAGE CXX)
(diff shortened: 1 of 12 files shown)


alk:
Thanks a lot! We could in principle consolidate do_mallinfo{,2} by making it template parameterized by int type we're using (size_t for "2" and int for "1"). But this is arguably good enough too.

alk:
Applied. Thanks a lot. And I amended your work by a change that consolidated those 2 functions as noted above.

PR #1467, opened 2023-12-16
mmap_hook.cc: use MAP_ANON when MAP_ANONYMOUS is not defined

barracuda156:
Older macOS (and possibly some other platforms) do not have `MAP_ANONYMOUS`, but have `MAP_ANON`.

### DIFF
--- a/src/mmap_hook.cc
+++ b/src/mmap_hook.cc
@@ -48,6 +48,9 @@
 #ifdef HAVE_MMAP
 # define mremap glibc_mremap
 # include <sys/mman.h>
+# ifndef MAP_ANONYMOUS
+#  define MAP_ANONYMOUS MAP_ANON
+# endif
 #include <sys/types.h>
 # undef mremap
 #endif


alk:
Thanks, but can you please move this under HAVE_MMAP define check and after we've included sys/mman.h ?

barracuda156:
@alk Is it correct now?

alk:
applied. thanks.

barracuda156:
@alk I overlooked last time because built it without tests. There are two more instances where the fix is due:

https://github.com/gperftools/gperftools/blob/8987d08f792c1c694849a6e779735bfcf50ac8a4/src/tests/check_address_test.cc#L49 (this for sure)

https://github.com/gperftools/gperftools/blob/8987d08f792c1c694849a6e779735bfcf50ac8a4/src/check_address-inl.h#L110

alk:
Thanks. You want to send a patch ?

PR #1470, opened 2023-12-24
fix setting page size and page size shift in cmake

### LINKED ISSUE #1373: cmake build seemingly lacks 4k pages option (was: Page size 4 kB - how to set?)
How do I set a page size of 4 kB? I can see there are at least two different build switches that I believe refers to the same thing? 

There is a parameter called `TCMALLOC_PAGE_SIZE_SHIFT`  that seems to allow me to change to 4 kB (shift of 12) 
https://github.com/gperftools/gperftools/blob/master/cmake/config.h.in#L284

The top cmake recipe has a parameter called `gperftools_tcmalloc_pagesize` but that only allows setting a page size down to 8 kB.
https://github.com/gperftools/gperftools/blob/master/CMakeLists.txt#L122

What is the correct way to set 4 kB page size?

m-fila:
Fixed cmake to allow the same page sizes as autotools, and to set `TCMALLOC_PAGE_SIZE_SHIFT` correctly.  
Removed setting `TCMALLOC_${gperftools_tcmalloc_pagesize}K_PAGES` that was never used.

Closes #1373

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -122,18 +122,31 @@ set(enable_libunwind ${gperftools_enable_libunwind})
 
 set(gperftools_tcmalloc_pagesize ${default_tcmalloc_pagesize}
   CACHE STRING "Set the tcmalloc internal page size")
-set_property(CACHE gperftools_tcmalloc_pagesize PROPERTY STRINGS "8" "32" "64")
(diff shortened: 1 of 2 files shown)


alk:
Nice. Thanks. Applied.

PR #1473, opened 2024-01-02
fix cmake pthread complile

qqwangxiaow:
I think we need to add a dependency on pthread here.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1023,7 +1023,7 @@ if(GPERFTOOLS_BUILD_DEBUGALLOC)
     endif()
 
     add_executable(realloc_debug_unittest ${realloc_unittest_SOURCES})
-    target_link_libraries(realloc_debug_unittest PUBLIC tcmalloc_minimal_debug)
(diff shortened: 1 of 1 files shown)


alk:
Hi. Thanks for the patch. But can you please elaborate what is the issue exactly that you're fixing here. What doesn't work without this patch ?

qqwangxiaow:
> Hi. Thanks for the patch. But can you please elaborate what is the issue exactly that you're fixing here. What doesn't work without this patch ?

Description:	Debian GNU/Linux 9.8 (stretch)
Release:	9.8
gcc version 9.4.0

I have encountered this problem
make -j 8
image

alk:
Thanks. it is clear now. But adding -lpthread is not the right fix. Some OSes don't have -lpthread. So I'll see what might be wrong.

P.S. Also for the later, please, consider learning how to quote regular text rather then posting screenshoots. Not as usable on mobile.

alk:
See my commit linked above. Should be fixed now.

Issue #1477, opened 2024-01-09
[macOS] `gperftools` 2.15: some tests fail


barracuda156:
I have ran tests now on 14.2.1 aarch64 with clang and on 10.6 ppc with gcc. There are a number of failures:

**14.2.1/aarch64**
```
/opt/local/bin/ctest --force-new-ctest-process 
Test project /opt/local/var/macports/build/_opt_svacchanda_SonomaPorts_devel_gperftools/gperftools/work/build
      Start  1: low_level_alloc_unittest
 1/46 Test  #1: low_level_alloc_unittest .............................   Passed    1.33 sec
      Start  2: stacktrace_unittest
 2/46 Test  #2: stacktrace_unittest ..................................   Passed    0.23 sec
      Start  3: check_address_unittest
 3/46 Test  #3: check_address_unittest ...............................   Passed    0.35 sec
      Start  4: pprof_unittest
 4/46 Test  #4: pprof_unittest .......................................   Passed    0.40 sec
      Start  5: tcmalloc_minimal_unittest
 5/46 Test  #5: tcmalloc_minimal_unittest ............................Subprocess aborted***Exception:   0.68 sec
      Start  6: tcm_min_asserts_unittest
 6/46 Test  #6: tcm_min_asserts_unittest .............................Subprocess aborted***Exception:   0.23 sec
      Start  7: tcmalloc_minimal_large_unittest
 7/46 Test  #7: tcmalloc_minimal_large_unittest ......................   Passed    0.25 sec
      Start  8: tcmalloc_minimal_large_heap_fragmentation_unittest
 8/46 Test  #8: tcmalloc_minimal_large_heap_fragmentation_unittest ...Subprocess aborted***Exception:   0.31 sec
      Start  9: addressmap_unittest
 9/46 Test  #9: addressmap_unittest ..................................   Passed    1.04 sec
      Start 10: system_alloc_unittest
10/46 Test #10: system_alloc_unittest ................................   Passed    0.23 sec
      Start 11: unique_path_unittest
11/46 Test #11: unique_path_unittest .................................   Passed    0.22 sec
      Start 12: packed_cache_test
12/46 Test #12: packed_cache_test ....................................   Passed    0.23 sec
      Start 13: frag_unittest
13/46 Test #13: frag_unittest ........................................   Passed    0.26 sec
      Start 14: markidle_unittest
14/46 Test #14: markidle_unittest ....................................   Passed    0.24 sec
[...]

Issue #1422, opened 2023-08-28
2.12.0: Tests fail if building with CMake


Lastique:
Here's the build log:

[cmake_fail.txt](https://github.com/gperftools/gperftools/files/12454164/cmake_fail.txt)

The last test is hanging, after a few minutes I interrupted it.

This is on Kubuntu 22.04. On the same machine, gperftools builds and tests pass if using automake.

alk:
Yes, our cmake support is incomplete and best-effort (which in practice is no effort almost)

PR #1478, opened 2024-01-10
Fix cmake tests

m-fila:
Some of the tests built with cmake were failing while the same test built with autotools were passing

- Fixed naming of cmake tests to easly correlate with autotools tests.
- Fixed broken paths to test executables.
- Fixed defining project version patch - `tcmalloc_unittest` and `tcmalloc_debug_unittest`

Related issues: #1422, #1477

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,7 @@
 /binary_trees.exe
 /binary_trees_shared
 /binary_trees_shared.exe
+/build
 /check_address_unittest
 /compile
 /config.guess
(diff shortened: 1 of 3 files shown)


alk:
Thanks for the patches. Looks fine. But somehow github is telling me this is "draft". Not sure what this means and whether it is intentional or not. Just let me know if you want me to merge those.

m-fila:
Thanks. This is a bit of a work in progress as I hope to add a few more fixes for the remaining tests relatively soon.

alk:
BTW, please consider this thought: I don't think it is good idea for us to try and to 100% match autotools-based features.

For example, I.e. am seeing that cmake lacks proper support for libtool-like feature of being able to build both libtcmalloc.so and libtcmalloc.a. Instead we do have the ugly libtcmalloc_static.a thingy. Perhaps we simply let users choose which *one* option to build via BUILD_SHARED_LIBS option.

And same likely applies elsewhere (we do try in autotools make file to test "with asserts" config, which we don't have to duplicate in cmake etc).

I.e. just give people simplest way to do cmake-way of building (most or all of) gperftools. Perhaps because they love cmake, and/or perphaps because cmake has IDE integrations. etc.

So feel free to reach out if you want to discuss direction.

alk:
BTW have a look at https://github.com/gperftools/gperftools/commit/a02ac30bbfb698992c2d0e9cd4edd5f7bd2de053

I intend cmake bits to get into similar simplified shape (actually even more). We're currently way too complicated there.

PR #1479, opened 2024-01-12
feat: make profile buffer size adjustable through env variable

### LINKED ISSUE #1450: Some allocation records disappear in between two consecutive dumps with -alloc_space or -alloc_objects
Hello, we see the following strange behavior with gperftools v2.9.1 as well as latest v2.13.

In a dump `0001.heap`, processed with `-alloc_space` (and `-top`) pprof options, we get the following output:

```
Type: alloc_space
Showing nodes accounting for 37059.20kB, 96.79% of 38287.35kB total
Dropped 1209 nodes (cum <= 191.44kB)
      flat  flat%   sum%        cum   cum%
29917.87kB 78.14% 78.14% 29917.87kB 78.14%  s1::util::StdNewDeleteResource::Allocate
         ... other items ...
         0     0% 96.79%  9625.52kB 25.14%  TiXmlAttribute::operator new
         0     0% 96.79% 25661.21kB 67.02%  TiXmlDocument::LoadFile
         0     0% 96.79% 25657.21kB 67.01%  TiXmlDocument::Parse
         0     0% 96.79% 27344.76kB 71.42%  TiXmlDocument::allocate
         0     0% 96.79% 26500.98kB 69.22%  TiXmlElement::Parse
         0     0% 96.79% 26063.46kB 68.07%  TiXmlElement::ReadValue
         0     0% 96.79% 17719.24kB 46.28%  TiXmlElement::operator new
         ... other items ...
```

We can see that there were allocations of ~ 38 MB, most of that coming from `TiXml*` classes through `s1::util::StdNewDeleteResource::Allocate`. However, in a later dump `0002.heap`, taken after the `TiXml` classes have released their memory, and again processed with `-alloc_space` option, we get the following output:

```
Type: alloc_space
Showing nodes accounting for 12040.94kB, 95.46% of 12613.90kB total
Dropped 1271 nodes (cum <= 63.07kB)
      flat  flat%   sum%        cum   cum%
 3418.61kB 27.10% 27.10%  3418.61kB 27.10%  s1::util::StdNewDeleteResource::Allocate
         ... other items ...
```

We see the absolute number of memory allocated dropped to ~ 12 MB. And no trace of `TiXml*` classes. How is that possible? Shouldn't `-alloc_space` and `-alloc_objects` numbers be ever-increasing with consecutive dumps, and catch all allocations performed until that point (including -- and this is essential here -- those that were deallocated)?

This appears to be a bug, and makes it impossible for us to use gperftools reliably, because we don't know whether it catches all allocation-deallocation peaks done in between two dumps.

What do you think the cause could be? Thanks much for your help.

Architecture we are using: aarch64

May be related to #1388

sangelovic:
The profile buffer size has until now been hard-coded to 1 MB. This led to a lot of lost data in our application profiling. We need at least 20 MB buffer size for our use case in order to get complete, trustworthy allocation data.

We have to patch gperftools in our build environment for that. But this is not a preferred solution. The preferred solution is that gperftools has this size configurable by clients at run-time. We guess more clients, not just us, may benefit from this possibility.

This PR adds a new `HEAP_PROFILE_BUFFER_SIZE` env variable for that purpose.

Fixes #1450

### DIFF
--- a/docs/heapprofile.html
+++ b/docs/heapprofile.html
@@ -158,6 +158,15 @@ environment variables.</p>
   </td>
 </tr>
 
+<tr valign=top>
+  <td><code>HEAP_PROFILE_BUFFER_SIZE</code></td>
+  <td>default: 1048576 (1 Mb)</td>
+  <td>Size of the buffer for heap profiling data (i.e., maximum size of the profile file).
+      Allocation samples beyond this limit will get stripped off. Applications with large
+      number of (de)allocations may need to increase this size to get complete, unstripped data.
+  </td>
+</tr>
+
 </table>
 
 <H2>Checking for Leaks</H2>
--- a/src/heap-profiler.cc
+++ b/src/heap-profiler.cc
@@ -120,6 +120,11 @@ DEFINE_bool(only_mmap_profile,
             EnvToBool("HEAP_PROFILE_ONLY_MMAP", false),
             "If heap-profiling is on, only profile mmap, mremap, and sbrk; "
             "do not profile malloc/new/etc");
+DEFINE_int32(profile_buffer_size,
+             EnvToInt64("HEAP_PROFILE_BUFFER_SIZE", 1 << 20 /*1MB*/),
+             "Size of buffer for heap profiling data. Applications with "
+             "large number of (de)allocations may need more than 1 MB "
+             "to avoid stripping off excess profiling data.");
 
 
 //----------------------------------------------------------------------
@@ -148,7 +153,7 @@ static void ProfilerFree(void* p) {
 }
 
 // We use buffers of this size in DoGetHeapProfile.
-static const int kProfileBufferSize = 1 << 20;
+static const int kProfileBufferSize = FLAGS_profile_buffer_size;
 
 // This is a last-ditch buffer we use in DumpProfileLocked in case we
 // can't allocate more memory from ProfilerMalloc.  We expect this


alk:
thanks for the patch. It looks fine. But we could do better.

There are 2 things we may consider. a) I think sorting of buckets isn't really necessary (pprof should sort and even de-dup itself); but I didn't double check. b) we can at runtime figure out right size (or double size in the loop until it fits), allocating memory with mmap (tcmalloc::DirectAnonMMap). This approach has a slight risk of OOMing whatever machine/container runs the process, but risks are really minimal (set of distinct backtraces that allocate memory tends to be quite bounded; probably why original static limit of 1 meg was put in place)

 Perhaps you have time for some more serious update? (if not, then, fine, I'll take this patch as-is)

alk:
And also if can find ways to avoid shell in our tests, that would be super-useful too. More portable and would also hopefully simplify cmake story too. I.e. consider simplifying everything rather than make cmake bits even more ugly to try to match existing complexity.

sangelovic:
@alk Please see https://github.com/gperftools/gperftools/pull/1481/files

alk:
Thanks, again, for your patch. On closer look I realized that I really want to make it much better than just plugging relatively minor details. So I went for more elaborate fix which I just merged. It should fix your issue and more.

I'd be thankful if you give those couple commits a look (for a chance that my self-review wasn't thorough enough).

PR #1481, opened 2024-01-15
Feat/make profile buffer size dynamic

sangelovic:
This is an alternative solution to https://github.com/gperftools/gperftools/pull/1479.

Profile data buffer size is no more fixed to 1 MB. The profile buffer starts at 1 MB size and grows exponentially dynamically until the profile data fits completely without stripping into the buffer.

### DIFF
--- a/src/heap-profile-table.cc
+++ b/src/heap-profile-table.cc
@@ -340,54 +340,42 @@ std::pair<int, bool> HeapProfileTable::FillOrderedProfile(char buf[], int size) 
   BucketsPtr list = MakeSortedBucketList();
 
   // Our file format is "bucket, bucket, ..., bucket, proc_self_maps_info".
-  // TODO: update comment
-  // In the cases buf is too small, we'd rather leave out the last
-  // buckets than leave out the /proc/self/maps info.  To ensure that,
-  // we actually print the /proc/self/maps info first, then move it to
-  // the end of the buffer, then write the bucket info into whatever
-  // is remaining, and then move the maps info one last time to close
-  // any gaps.  Whew!
-  int map_length = snprintf(buf, size, "%s", kProcSelfMapsHeader);
-  if (map_length < 0 || map_length >= size) return {0, false};
-
-  bool wrote_all{}; // "wrote_all" -- did /proc/self/maps fit in its entirety?
-  map_length += FillProcSelfMaps(buf + map_length, size - map_length, &wrote_all);
-  RAW_DCHECK(map_length <= size, "");
-  if (!wrote_all) return {map_length, false};
-
-  char* const map_start = buf + size - map_length;      // move to end
-  memmove(map_start, buf, map_length);
-  size -= map_length;
 
   Stats stats;
   memset(&stats, 0, sizeof(stats));
-  int bucket_length = snprintf(buf, size, "%s", kProfileHeader);
-  if (bucket_length < 0 || bucket_length >= size) return {0, false};
+  int filled_length = snprintf(buf, size, "%s", kProfileHeader);
+  if (filled_length < 0 || filled_length >= size) return {0, false};
 
-  std::tie(bucket_length, wrote_all) = UnparseBucket(total_, buf, bucket_length, size,
+  bool wrote_all{}; // Did dumped buckets and maps fit in its entirety?
+  std::tie(filled_length, wrote_all) = UnparseBucket(total_, buf, filled_length, size,
                                                      " heapprofile", &stats);
   if (!wrote_all) return {0, false};
 
   // Dump the mmap list first.
   if (profile_mmap_) {
-    BufferArgs buffer(buf, bucket_length, size);
+    BufferArgs buffer(buf, filled_length, size);
     MemoryRegionMap::LockHolder holder{};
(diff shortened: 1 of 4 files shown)


sangelovic:
@alk What do you think about this one? I don't have much time and no real insight into gperftools code, but I gave this one a chance, and implemented a dynamically growing buffer from the list of your suggestions. Please review.

I used `ProfilerMalloc` to allocate the buffer (it was used in original code), as opposed to your suggested `DirectAnonMMap`. But if `DirectAnonMMap` is a better fit, I'll change that.

A few functions return an extra `wrote_all` or `complete` boolean flag, indicating whether all data in the call got written into the buffer.

I also did a little modernization -- I used `unique_ptr` for bucket resource ownership, which helped to make the code less bug-prone (many `dealloc_` calls disappeared).

One thing that is possible with this change is that no more ProcSelfMaps have to be serialized first, then buckets in front of it. Now we can serialize them in order as they appear in the dump file (first buckets, then maps). I can simplify this, too, but let me know what you think.

alk:
Thanks. This is great. How about the following small update to your change:

```
diff --git a/src/heap-profile-table.cc b/src/heap-profile-table.cc
index 186519c4..82d8d0b6 100644
--- a/src/heap-profile-table.cc
+++ b/src/heap-profile-table.cc
@@ -308,8 +308,7 @@ std::pair<int, bool> HeapProfileTable::UnparseBucket(const Bucket& b,
   return {buflen, true};
 }
 
-std::unique_ptr<HeapProfileTable::Bucket*, HeapProfileTable::DeAllocator>
-HeapProfileTable::MakeSortedBucketList() const {
+HeapProfileTable::BucketsPtr HeapProfileTable::MakeSortedBucketList() const {
   Bucket** list = static_cast<Bucket**>(alloc_(sizeof(Bucket) * num_buckets_));
 
   int bucket_count = 0;
@@ -330,15 +329,15 @@ void HeapProfileTable::IterateOrderedAllocContexts(
   auto list = MakeSortedBucketList();
   AllocContextInfo info;
   for (int i = 0; i < num_buckets_; ++i) {
-    *static_cast<Stats*>(&info) = *static_cast<Stats*>(list.get()[i]);
-    info.stack_depth = list.get()[i]->depth;
-    info.call_stack = list.get()[i]->stack;
+    *static_cast<Stats*>(&info) = *static_cast<Stats*>(list[i]);
+    info.stack_depth = list[i]->depth;
+    info.call_stack = list[i]->stack;
     callback(info);
   }
 }
 
 std::pair<int, bool> HeapProfileTable::FillOrderedProfile(char buf[], int size) const {
-  auto list = MakeSortedBucketList();
+  BucketsPtr list = MakeSortedBucketList();
 
   // Our file format is "bucket, bucket, ..., bucket, proc_self_maps_info".
   // TODO: update comment
@@ -379,7 +378,7 @@ std::pair<int, bool> HeapProfileTable::FillOrderedProfile(char buf[], int size)
   }
 
   for (int i = 0; i < num_buckets_; i++) {
-    std::tie(bucket_length, wrote_all) = UnparseBucket(*list.get()[i], buf, bucket_length, size, "",
+    std::tie(bucket_length, wrote_all) = UnparseBucket(*list[i], buf, bucket_length, size, "",
                                                        &stats);
     if (!wrote_all) return {0, false};
   }
diff --git a/src/heap-profile-table.h b/src/heap-profile-table.h
index 22748a1c..31e67f91 100644
--- a/src/heap-profile-table.h
+++ b/src/heap-profile-table.h
@@ -189,6 +189,7 @@ class HeapProfileTable {
   // Hash table bucket to hold (de)allocation stats
   // for a given allocation call stack trace.
   typedef HeapProfileBucket Bucket;
+  typedef std::unique_ptr<Bucket*[], DeAllocator> BucketsPtr;
 
   // Info stored in the address map
   struct AllocValue {
@@ -302,7 +303,7 @@ class HeapProfileTable {
   // Helper for IterateOrderedAllocContexts and FillOrderedProfile.
   // Creates a sorted list of Buckets whose length is num_buckets_.
   // The caller is responsible for deallocating the returned list.
-  std::unique_ptr<HeapProfileTable::Bucket*, HeapProfileTable::DeAllocator> MakeSortedBucketList() const;
+  BucketsPtr MakeSortedBucketList() const;
 
   // Helper for TakeSnapshot.  Saves object to snapshot.
   static void AddToSnapshot(const void* ptr, AllocValue* v, Snapshot* s);
```

Also, I don't think there is much value keeping 2 commits. Can you squash? And also please take care of remaining TODO item you've added.

alk:
Ah, just spotted your last statement about order of serialization. Sure, go ahead and try that as well (this one is fine to try in distinct commit).

sangelovic:
@alk Incorporated your patch (I like it, using array-based `unique_ptr` is a better fit here), re-ordered the serializations so how they are serialized in code is also how they appear in the dump file, and squashed commits as desired.

alk:
Thanks, again, for your patch. On closer look I realized that I really want to make it much better than just plugging relatively minor details. So I went for more elaborate fix which I just merged. It should fix your issue and more.

I'd be thankful if you give those couple commits a look (for a chance that my self-review wasn't thorough enough).

PR #1482, opened 2024-01-16
Set NOMINMAX in config.h.in to unblock builds with more recent versio…

lennoxho:
I noticed MSVC CMake builds are broken with a more recent copy of the Windows SDK.

Quick and easy fix - set NOMINMAX in config.h.in to unblock builds with more recent versions of Windows SDK.

### DIFF
--- a/cmake/config.h.in
+++ b/cmake/config.h.in
@@ -7,6 +7,10 @@
 #define GOOGLE_PERFTOOLS_WINDOWS_CONFIG_H_
 #endif
 
+#if defined(_WIN32) && !defined(NOMINMAX)
+#define NOMINMAX
+#endif
+
 #ifndef GOOGLE_PERFTOOLS_WINDOWS_CONFIG_H_
 #define GOOGLE_PERFTOOLS_WINDOWS_CONFIG_H_
 /* used by tcmalloc.h */


lennoxho:
Absorbed into https://github.com/gperftools/gperftools/pull/1485

alk:
We shouldn't be doing this in cmake-specific config.h. We already have this in windows' generic src/windows/config.h why isn't it working? (if not then we should find a way to do the more portably; cannot afford diverging cmake/autotools/msbuild stuff)

lennoxho:
So the CMake build pulls in its own `config.h` in some cases, `src/windows/config.h` in other cases (depends on whether it's building source files from `src/windows`).

It's quite inconsistent at the moment, and this causes conflicts in the CMake MinGW build (as the contents from `src/windows/config.h` does not exactly match what we managed to deduce in CMake).

In https://github.com/gperftools/gperftools/pull/1485 which supersedes this PR, I'm attempting to remove this inconsistency by isolating the `config.h` used by VSProj/Autotools and the `config.h` used in the CMake builds.

Totally agree with the issue of potentially diverging `config.h`, it's messy.
I think a long term solution would be to agree on a common build system for all supported platforms.

Until we settle on what that would look like, hopefully we can apply the proposed patches to unblock the MSVC & MinGW builds.

alk:
No defines in cmake-specific bits, please. We should aim to do this in ports.h or similar.

Issue #1486, opened 2024-01-16
MinGW Autotools & CMake builds are broken


lennoxho:
Autools MSYS [link](https://github.com/msys2/msys2-installer/releases/tag/2024-01-13) build:
```
g++ -std=gnu++11 -DHAVE_CONFIG_H -I. -I./src  -I./src -DNO_TCMALLOC_SAMPLES  -pthread -Wall -Wwrite-strings -Woverloaded-virtual -Wno-sign-compare -Wno-unused-result -fsized-deallocation -faligned-new  -fno-omit-frame-pointer -momit-leaf-frame-pointer -g -O2 -MT src/tests/page_heap_test-page_heap_test.o -MD -MP -MF src/tests/.deps/page_heap_test-page_heap_test.Tpo -c -o src/tests/page_heap_test-page_heap_test.o `test -f 'src/tests/page_heap_test.cc' || echo './'`src/tests/page_heap_test.cc
In file included from D:/Users/LennoxHo/Downloads/New/msys64/ucrt64/include/c++/13.2.0/x86_64-w64-mingw32/bits/gthr-default.h:35,
                 from D:/Users/LennoxHo/Downloads/New/msys64/ucrt64/include/c++/13.2.0/x86_64-w64-mingw32/bits/gthr.h:148,
                 from D:/Users/LennoxHo/Downloads/New/msys64/ucrt64/include/c++/13.2.0/ext/atomicity.h:35,
                 from D:/Users/LennoxHo/Downloads/New/msys64/ucrt64/include/c++/13.2.0/bits/shared_ptr_base.h:61,
                 from D:/Users/LennoxHo/Downloads/New/msys64/ucrt64/include/c++/13.2.0/bits/shared_ptr.h:53,
                 from D:/Users/LennoxHo/Downloads/New/msys64/ucrt64/include/c++/13.2.0/memory:80,
                 from src/tests/page_heap_test.cc:13:
D:/Users/LennoxHo/Downloads/New/msys64/ucrt64/include/pthread.h:187:18: error: conflicting declaration 'typedef unsigned int pthread_key_t'
  187 | typedef unsigned pthread_key_t;
      |                  ^~~~~~~~~~~~~
In file included from ./src/windows/mingw.h:68,
                 from ./src/config.h:275,
                 from ./src/config_for_unittests.h:58,
                 from src/tests/page_heap_test.cc:8:
[...]

Issue #1483, opened 2024-01-16
TCMalloc built with mingw-w64 deadlocks during startup when pthread-based TLS is enabled


lennoxho:
Here's the call stack
![tcmalloc mingw pthread deadlock no tls](https://github.com/gperftools/gperftools/assets/9140100/deba1232-9e08-4585-9f79-c88217a90614)

The `??`s from `libwinpthread-1.dll` are calls to `pthread_setspecific()`.
Here's the line we are stuck on (it's a spinlock) https://github.com/mingw-w64/mingw-w64/blob/86a5e0f41683d8e60b3756712e7a417055d291dc/mingw-w64-libraries/winpthreads/src/thread.c#L913

The spinlock implementation in question
https://github.com/mingw-w64/mingw-w64/blob/86a5e0f41683d8e60b3756712e7a417055d291dc/mingw-w64-libraries/winpthreads/src/spinlock.c#L50

Looks like the MinGW port of `pthread_setspecific()` is not reentrant.

Identified with this flavour of MinGW
https://github.com/brechtsanders/winlibs_mingw/releases/tag/13.2.0posix-17.0.5-11.0.1-ucrt-r3

The build flag in question is `HAVE_PTHREAD`.
I'm currently working on a changelist to disable it by default in the CMake build

lennoxho:
https://github.com/gperftools/gperftools/pull/1485

alk:
First of all, thanks a lot for looking at this and reporting the issue.

So I am able to reproduce the deadlock and indeed the issue is they're calling realloc while holding their internal lock.

Deadlock is only reproducible on "posix" variant of mingw that debian ships with mingw cross compiler, not their default "win32" variant. (And this is all mega-stale msvcrt, sadly no ucrt option there yet)

Here is what I think we should be doing:

a) general philosophy is now to embrace threading rather than trying to avoid it. So all this "don't include pthreads" logic should go. It only adds complications and multiples configurations that require additional testing. (It made sense for us in past, but not anymore; see #1110 for some more background).

[...]

Issue #1484, opened 2024-01-16
TCMalloc built with mingw-w64 encounters infinite recursion during startup when implicit TLS is enabled


lennoxho:
The call stack:
![tcmalloc mingw pthread deadlock tls](https://github.com/gperftools/gperftools/assets/9140100/9c858d4d-b83a-41ac-8ec8-5f9570999005)

Note the 2 `__emutls_get_address` calls shown.
__emutls_get_address calls `malloc` during intialisation
https://github.com/gcc-mirror/gcc/blob/57f611604e8bab67af6c0bcfe6ea88c001408412/libgcc/emutls.c#L153

GCC does not (yet) support true implicit TLS on Windows.
It instead generates `__emutls_get_address` (port of `__tls_get_address`)... and I guess it's not reentrant too.

Identified with this flavour of MinGW
https://github.com/brechtsanders/winlibs_mingw/releases/tag/13.2.0posix-17.0.5-11.0.1-ucrt-r3

The build flag in question is HAVE_TLS.
It's appears to be disabled by default in both the CMake and MinGW builds. So I would say it's low visibility.

lennoxho:
Will be addressed as described here https://github.com/gperftools/gperftools/issues/1483#issuecomment-1894500594

lennoxho:
As noted, this issue is not visible unless HAVE_TLS is explicitly set to ON with MinGW builds.

Closing...

PR #1487, opened 2024-01-16
Fix MinGW deadlock when using the POSIX thread model; Fix MinGW Autotools build

lennoxho:
Hello!

This is the 1st of two PRs described here
https://github.com/gperftools/gperftools/issues/1483#issuecomment-1894500594

This PR fixes the following MinGW issues:
1. `pthread_*` redefinition errors in `page_heap_test` as described in #1486
2. Deadlock in `pthread_getspecific` as described in #1483

Both issues are resolved by changing the MinGW build to always rely on the native WinAPI for threading support.

**Testing done:**

Autotools:
1. WSL Ubuntu 22.04 LTS + GCC 13.1
2. MSYS2-MinGW [link](https://github.com/msys2/msys2-installer/releases/tag/2024-01-13) + GCC 13.2

VSProj:
1. MSVC 19.38 + WinSDK 10.0.22621.0

CMake:
1. WSL Ubuntu 22.04 LTS + GCC 13.1
2. MSVC 19.38 + WinSDK 10.0.22621.0
3. **[Still Broken, will be addressed in 2nd PR]** MinGW [link](https://github.com/brechtsanders/winlibs_mingw/releases/tag/13.2.0posix-17.0.5-11.0.1-ucrt-r3) + GCC 13.2

### DIFF
--- a/src/base/threading.h
+++ b/src/base/threading.h
@@ -1,0 +1,127 @@
+// -*- Mode: C++; c-basic-offset: 2; indent-tabs-mode: nil -*-
+// Copyright (c) 2024, gperftools Contributors
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are
+// met:
+//
+//     * Redistributions of source code must retain the above copyright
+// notice, this list of conditions and the following disclaimer.
+//     * Redistributions in binary form must reproduce the above
+// copyright notice, this list of conditions and the following disclaimer
+// in the documentation and/or other materials provided with the
+// distribution.
+//     * Neither the name of Google Inc. nor the names of its
+// contributors may be used to endorse or promote products derived from
+// this software without specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+// "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+// LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+// A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+// OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+// SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+// LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+// DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+// THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+// (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+#ifndef THREADING_H_
+#define THREADING_H_
+
+#include <config.h>
+#include "base/basictypes.h"
+
+// Also allow for printing of a PerftoolsThreadID.
+#define GPRIuPTHREAD "lu"
+#define GPRIxPTHREAD "lx"
+
+#if defined(__CYGWIN__) || defined(__CYGWIN32__) || defined(__APPLE__) || defined(__FreeBSD__)
+#  define PRINTABLE_PTHREAD(pthreadt) reinterpret_cast<uintptr_t>(pthreadt)
+#elif defined(__QNXNTO__)
+#  define PRINTABLE_PTHREAD(pthreadt) static_cast<intptr_t>(pthreadt)
+#else
+#  define PRINTABLE_PTHREAD(pthreadt) pthreadt
+#endif
+
+#ifdef _WIN32 // Should cover both toolchains on Windows - MSVC & MINGW
+
+using PerftoolsThreadID = DWORD;
+using PerftoolsTlsKey = DWORD;
+
+inline PerftoolsThreadID PerftoolsGetThreadID() {
+  return GetCurrentThreadId();
+}
+inline int PerftoolsThreadIDEquals(PerftoolsThreadID left, PerftoolsThreadID right) {
+  return left == right;
+}
+
+extern "C" PerftoolsTlsKey PthreadKeyCreate(void (*destr_fn)(void*));  /* port.cc */
+
+inline int PerftoolsCreateTlsKey(PerftoolsTlsKey *pkey, void (*destructor)(void*)) {
+  PerftoolsTlsKey key = PthreadKeyCreate(destructor);
+
+  if (key != TLS_OUT_OF_INDEXES) {
+    *(pkey) = key;
+    return 0;
+  }
+  else {
+    return GetLastError();
+  }
+}
+inline void* PerftoolsGetTlsValue(PerftoolsTlsKey key) {
+  DWORD err = GetLastError();
+  void* rv = TlsGetValue(key);
+
+  if (err) SetLastError(err);
+  return rv;
+}
+inline int PerftoolsSetTlsValue(PerftoolsTlsKey key, const void* value) {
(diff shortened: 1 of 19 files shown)


src/windows/port.h
```diff
@@ -108,61 +108,6 @@ typedef intptr_t ssize_t;
 
 /* ----------------------------------- THREADS */
 
-#ifndef HAVE_PTHREAD   /* not true for MSVC, but may be true for MSYS */
```
lennoxho, line 111: Moved to `base/threading.h`

src/base/threading.h
```diff
@@ -0,0 +1,115 @@
...
+// (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+#ifndef THREADING_H_
+#define THREADING_H_
+
+#include <config.h>
+#include "base/basictypes.h"
+
+#ifdef _WIN32 // Should cover both toolchains on Windows - MSVC & MINGW
```
lennoxho, line 37: Previously we were predicating on `HAVE_PTHREAD`.
Switching to `_WIN32` is how we redirect MinGW builds away from `pthread`s and use `WinAPI` instead.

alk:
Thanks. Yes, this is the direction I want. Only one thing, please don't replace sched_yield calls in Linux or POSIX-specific codes (like spinlock or heap checker stuff).

lennoxho:
> please don't replace sched_yield calls in Linux or POSIX-specific codes (like spinlock or heap checker stuff).

Done

alk:
Hi. This is still great. You mentioned above that you intend to give it some extra testing ? Or is it ready to be merged ?

lennoxho:
> Hi. This is still great. You mentioned above that you intend to give it some extra testing ? Or is it ready to be merged ?

Hi!

I've tested this PR on all platforms available to me (listed above). Everything looks good.
Should be good to go.

Thanks!

alk:
Aha. Great. Then please squash commits and make proper commit message (i.e. it still says draft). Or just let me know and I'll squash and rewrite commit message myself (but I'd definitely prefer you do it).

lennoxho:
> please squash commits and make proper commit message

Alright. Done!
Also sanity checked to make sure I didn't break anything while squashing.

alk:
Ah, I missed couple things.

* I see you added copyright header. This is right. But it states copyright google inc, which isn't right. We're putting "gperftools Contributors" everywhere. So please do so as well. (Rest I can do as a follow-up, but this one, we have to get right on original commit; again, I can amend myself as part of merging, but I prefer you do it)

* HAVE_PTHREAD isn't useful anymore. We simply assume it is pthreads if it is not windows.
* couple tests do pthread APIs directly, e.g. to create threads, which is fine (to be handler later and separately). But then please don't mix new custom types and functions in there (testutil.cc, profile-handler_unittest.cc, heap-checker_unittest.cc), since it mixes things up and adds confusion.
* PRINTABLE_PTHREAD is worth bringing into threading.h somehow. Notably, we aren't even using it consistently. So do we really need it is one question.
* PerftoolsThreadIDEquals is only used to test if current thread's id is same as given thread id. So maybe consider having such function directly rather than "equals". Another thought is should we consider std::thread::id instead (https://en.cppreference.com/w/cpp/thread/thread/id). Or are you saying mingw+pthread config breaks it?
* Similarly is it too bold to try doing TLS cleanup via C++11 threading facility?
* naming thing. We'll eventually get to a proper -fdefault-visibility treatment but for now we should be extra careful w.r.t. adding new exported symbols. You already did Perftools prefix, so not bad. But I'd prefer tcmalloc:: namespace instead (again feel free to leave it and I'll handle separately). I also do ATTRIBUTE_VISIBILITY_HIDDEN for all new symbols, but tcmalloc:: namespace is kinda enough too.

lennoxho:
I've made those changes:

> * I see you added copyright header. This is right. But it states copyright google inc, which isn't right. We're putting "gperftools Contributors" everywhere. So please do so as well. (Rest I can do as a follow-up, but this one, we have to get right on original commit; again, I can amend myself as part of merging, but I prefer you do it)
> * PRINTABLE_PTHREAD is worth bringing into threading.h somehow. Notably, we aren't even using it consistently. So do we really need it is one question.

-----------------

> * Similarly is it too bold to try doing TLS cleanup via C++11 threading facility?

While C++11 introduced `thread_local`, we still don't have standardised "explicit" TLS functions to replace `pthread_setspecific` and `TlsSetValue`.

Furthermore, `GCC` is known to produce different code for `__thread` vs `thread_local`
https://godbolt.org/z/5oGrME6qE

-----------------

I'll be happy to work on those in a follow up PR

> * HAVE_PTHREAD isn't useful anymore. We simply assume it is pthreads if it is not windows.
> * couple tests do pthread APIs directly, e.g. to create threads, which is fine (to be handler later and separately). But then please don't mix new custom types and functions in there (testutil.cc, profile-handler_unittest.cc, heap-checker_unittest.cc), since it mixes things up and adds confusion.
> * PerftoolsThreadIDEquals is only used to test if current thread's id is same as given thread id. So maybe consider having such function directly rather than "equals". Another thought is should we consider std::thread::id instead (https://en.cppreference.com/w/cpp/thread/thread/id). Or are you saying mingw+pthread config breaks it?
> * naming thing. We'll eventually get to a proper -fdefault-visibility treatment but for now we should be extra careful w.r.t. adding new exported symbols. You already did Perftools prefix, so not bad. But I'd prefer tcmalloc:: namespace instead (again feel free to leave it and I'll handle separately). I also do ATTRIBUTE_VISIBILITY_HIDDEN for all new symbols, but tcmalloc:: namespace is kinda enough too.

alk:
> Furthermore, GCC is known to produce different code for __thread vs thread_local https://godbolt.org/z/5oGrME6qE

This is due to external. If you remove that and the compiler sees that TLS variable initialization is "trivial" then it becomes identical. I.e. because C++ TLS has some special "features" required by standard, it then causes extra complications. But this only applies whenever initialization is nontrivial.

AFAIR even tls model attribute is supported too.

But I was referring to the fact that C++ standard TLS also natively supports running destructors for TLS variable instances. Which is how we're currently using pthread TLS stuff for platforms that define HAVE_TLS.

I.e. I'd consider seeking replacement for pthread_{get,set}specific by using those C++ TLS instances. Anyways, something I can do separately.

alk:
Wow. Thanks for catching the M_PI thingy. I wasn't aware it is non-standard...

alk:
merged. Thanks, again!

lennoxho:
Thanks!
I'll get the second PR ready by tomorrow.

Briefly here's what it's going to contain:
1) The 3 follow up items in the previous comments (I'll defer the thread_local changes to you)
2) Fix the Cmake MinGW & MSVC builds
3) I noticed some bad inconsistencies with config.h in the Cmake builds. It could potentially pull in one of 3 config.h headers depending on the source file being compiled. I plan to completely segregate those to their respective builds, pending consolidation down the road.

lennoxho:
Following up on C++ TLS, 

> But this only applies whenever initialization is nontrivial.

Oh yeah, however IIRC this optimisation for thread_local was not consistent at least when I last tested it with GCC 7. If I'm not mistaken the compiler will lazy-init for anything other than fundamental types. I could be mistaken though.

> I'd consider seeking replacement for pthread_{get,set}specific by using those C++ TLS instances

Do you mean we want to enforce HAVE_TLS and require implicit TLS support from the toolchains?

As noted in #1434 this could be problematic for platforms without reentrant implicit TLS support

alk:
> Do you mean we want to enforce HAVE_TLS and require implicit TLS support from the toolchains?

No. When we deal with HAVE_TLS we still do fast TLS pointer-only thingy for fast-path access. And "slow" and destructor-ful TLS that deals with cleanup. When we deal with !HAVE_TLS then we only do "slow" TLS both to access thread's thread-cache pointer and cleanup. Basically whenever we currently do pthread thread-specific things, we replace with C++ "native" tls. It think it could be tried.

alk:
The risk though is mingw when it is configured for pthread's case, it might end up with same deadlock you originally reported.

lennoxho:
> And "slow" and destructor-ful TLS that deals with cleanup

Didn't notice that. I guess I have some more code reading to do

> The risk though is mingw when it is configured for pthread's case, it might end up with same deadlock you originally reported.

HAVE_TLS is forced OFF in both MinGW Autotools and CMake builds. That takes care of #1484 

Also, with this PR MinGW builds will always use WinAPI's explicit "slow" TLS functions even when pthread is enabled.

So I believe this shouldn't be an issue

PR #1489, opened 2024-01-25
Several build fixes and cleanups

lennoxho:
Follow up to #1487

This PR includes the following fixes:
1. Fix CMake builds for MinGW and MSVC #1486
2. Ensure the Autotools, CMake and VSProj builds do not reference each others' `config.h`
    - Note we already have 3 config.h files today. We are maintaining this status quo for now.
    - Instead, the changes here is aimed at making sure each build consistently includes `config.h` first while ensuring the right header is actually read during compilation.
    - Currently the MinGW Autotools build is redirected via a hack to the Autotools generated configuration file, while the Windows CMake build usually includes the copy generated by CMake, but sometimes includes the copy under `src/windows/`

As well as those cleanup tasks:
1. Use `std::thread::id` instead of our own thread ID wrappers
2. Moved explicit TLS wrapper functions into the `tcmalloc:: namespace` and change their visibility to `hidden`

-------------

Not addressed:
1. Remove and replace `HAVE_PTHREAD` with `!_WIN32`
    - Slightly risky, want to tackle this item in another follow up
2. Replace TLS usage with C++11's `thread_local` to take advantage of auto-cleanup
    - I'll defer this item to @alk 

Future work:
1. Somehow consolidate the Autotools, Cmake and VSProj `config.h` header files

-------------

Testing done:

Autotools:

1. WSL Ubuntu 22.04 LTS + GCC 13.1
4. MSYS2-MinGW [link](https://github.com/msys2/msys2-installer/releases/tag/2024-01-13) + GCC 13.2

VSProj:

1. MSVC 19.38 + WinSDK 10.0.22621.0

 
CMake:
1. WSL Ubuntu 22.04 LTS + GCC 13.1
2. MSVC 19.38 + WinSDK 10.0.22621.0
5. MinGW [link](https://github.com/brechtsanders/winlibs_mingw/releases/tag/13.2.0posix-17.0.5-11.0.1-ucrt-r3) + GCC 13.2


configure.ac
```diff
@@ -255,7 +255,10 @@ AC_CHECK_TYPES([Elf32_Versym],,, [#include <elf.h>])   # for vdso_support.h
 AC_CHECK_FUNCS(sbrk)            # for tcmalloc to get memory
 AC_CHECK_FUNCS(__sbrk)          # for tcmalloc to get memory
 AC_CHECK_FUNCS(geteuid)         # for turning off services when run as root
-AC_CHECK_FUNCS(fork)            # for the pthread_atfork setup
+# for the pthread_atfork setup
+case "$host" in
+   *-linux*) AC_CHECK_FUNCS(fork);;
```
alk, line 260: sorry but this isn't right. Lots more OSes have fork (e.g. osx, bsd-s, qnx etc). Main question is why? I think intention you have is to disable fork testing on mingw, but why?

configure.ac
```diff
@@ -541,6 +544,12 @@ AC_SUBST(LIBSTDCXX_LA_LINKER_FLAG)
 # In fact, a lot of the code in this directory depends on pthreads
 AX_PTHREAD
 
+AC_LINK_IFELSE([AC_LANG_PROGRAM([#include <sys/types.h>],
```
alk, line 547: why? Can we simply do "windows doesn't have it, everyone else has" ? Because in practice it'll more or less like that.

src/base/threading.h
```diff
@@ -1,127 +1,110 @@
-// -*- Mode: C++; c-basic-offset: 2; indent-tabs-mode: nil -*-
```
alk, line 1: hm, what happened here? Line endings?

alk:
Thanks for your work. Main question I have is can you please elaborate why it is necessary to remove ```#include "config.h"``` from everywhere ? Because having makefiles, etc adding CPPFLAG to autoinclude this header means we're relying on certain compiler flags to be supported. Yes, most times we're dealing either with msvc or gcc-compatible compilers (e.g. intel supports gcc flags on !windows configs, clangs does same etc). But there are still occasional compilers around with own odd flags (e.g. check the crap we test for in AX_PTHREAD)

alk:
If think you really think we have to split config.h files, then perhaps we keep trivial config.h that depending on single define includes either, say, config-ac.h, config-cmake.h, config-msbuild.h etc. I.e. autotools already arranges for -DHAVE_CONFIG_H

I think it is possible to override config.h file names used by cmake and autotools.

I am not really sure why it has to be required but we (or I) can sort this up as a follow up.

src/base/threading.h
lennoxho: Oops. Let me fix that.

configure.ac
lennoxho: Unfortunately MinGW may have `pid_t`. And it seems to be inconsistent between flavours of MinGW.

configure.ac
lennoxho: > I think intention you have is to disable fork testing on mingw

You are right. And good point RE other OSes.
Just figured out how to properly check `if !MINGW` in Automake.

I've updated this check.

As to why we want to disable `HAVE_FORK` for MinGW, this was migrated from `mingw.h`, where we force `#undef HAVE_FORK`.
I figured having this hack officially documented in Automake/CMake is more visible and least surprising to developers.

lennoxho:
> Main question I have is can you please elaborate why it is necessary to remove `#include "config.h"` from everywhere ?

So it's not strictly necessary to remove those includes.
I was going with the general philosophy of "don't repeat yourself".
I can revert those if you prefer we leave them alone.

> If you really think we have to split config.h files,

I think we all agree this is probably not going to be a long term solution.

In our previous conversations we were talking about either 
- Finally settling on **a** (singular) build tool, OR
- Somehow merging all `config.h` template files and have it magically work with both Autotools and CMake (I think for VSProj we have no choice but to deal with a pregenerated header).

alk:
Before everything else, we should prefer minimizing config.h dependencies. And even more so, ./configure-time checkings. Because those checkings are mess, especially when we're somehow resorting to "but don't test it on mingw" and what not, _and_ when we duplicate those between cmake and autotools.

For example why bother with pid_t at all. It is used by heap-checker stuff (which is Linux-only anyways). And it is only nominally used by proc-maps-iterator bits. But we always use it in "myself" mode, so pid isn't really used there.

Similarly, whatever stock check for fork did, we can always ifdef-out it's usage for mingw. Keeping complexity out of cmake bits.

That cmake requires us to have config.h.in "template" is sad, but it is what it is. For autoconf producing whatever number of defines in config.h is trivial. And maintaining hardcoded straightforward config.h for MSBuild isn't hard too. So IMHO we just keep what we have today.

In principle, I am potentially open for duplicating config.h-es. But I am quite certain we cannot afford dropping those #includes and relying on /FI or -include.

alk:
Let me do discard pid_t stuff right now. Everything to simplify things helps. I may even do fork bits compile-time too.

lennoxho:
> Let me do discard pid_t stuff right now.

Cool!

> But I am quite certain we cannot afford dropping those #includes

In the meantime do you want me to revert those? Could clean up the diff quite a bit.

alk:
> In the meantime do you want me to revert those? Could clean up the diff quite a bit.

Yes, please.

BTW pid_t stuff is merged. Have a look, please, if you have couple minutes.

lennoxho:
> pid_t 
> HAVE_FORK 

Awesome!

I've merged, squashed and reverted the `#include "config.h"` changes.
Let me know what you think.

vsprojects/common.vcxproj.props
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="Current" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <ItemDefinitionGroup>
+    <ClCompile>
+      <ForcedIncludeFiles>config.h</ForcedIncludeFiles>
```
alk, line 5: is that required ? What does it do ?

src/base/threading.h
```diff
@@ -33,92 +33,76 @@
...
-}
-
-inline void PerftoolsYield() {
-  Sleep(0);
-}
+namespace tcmalloc {
+
+    using TlsKey = DWORD;
+
+    extern "C" TlsKey PthreadKeyCreate(void (*destr_fn)(void*));  /* windows/port.cc */
```
alk, line 55: would be most sensible to remove extern "C" here.

CMakeLists.txt
```diff
@@ -476,6 +476,13 @@ configure_file(cmake/tcmalloc.h.in
                ${CMAKE_CURRENT_BINARY_DIR}/gperftools/tcmalloc.h
                @ONLY)
 
+# Always include config.h, and include it first.
+if(MSVC)
+   add_compile_options(/FI${CMAKE_CURRENT_BINARY_DIR}/config.h)
```
alk, line 481: my understanding is we're not doing this part anymore right ?

lennoxho: While moving `src/windows/config.h` out of `src/` already sufficiently resolve the ambiguity described in the PR,

There's still a potential conflict if a CMake build is performed on a repository that went through a automake build previously.

In this scenario, we'll have:
1. <CMake build dir>/config.h
2. src/config.h

This is an issue because both GCC and MSVC will prefer the header file relative to the source file which includes it.
That means (2), which is the stale automake generated config file will be consumed for source files directly under `src`, while (2) will be consumed for source files under the descendant directories.

vsprojects/common.vcxproj.props
lennoxho: This is equivalent to `/FI` on the command line, which force includes the header file before any other header files.

src/base/threading.h
lennoxho: Done!

CMakeLists.txt
lennoxho: The same problem can be observed if the same repo was used for a Linux automake build, followed by a VSProj build.

```
Error	C1189	#error:  Could not determine cache line length - unknown architecture	page_heap_test	D:\Users\LennoxHo\source\gperftools-main\src\base\basictypes.h	379		
Error	C1083	Cannot open include file: 'unistd.h': No such file or directory	page_heap_test	D:\Users\LennoxHo\source\gperftools-main\src\common.cc	38		
Error	C1189	#error:  Could not determine cache line length - unknown architecture	libtcmalloc_minimal	D:\Users\LennoxHo\source\gperftools-main\src\base\basictypes.h	379		
Error	C1083	Cannot open include file: 'unistd.h': No such file or directory	addressmap_unittest	D:\Users\LennoxHo\source\gperftools-main\src\base\logging.h	44		
Error	C1083	Cannot open include file: 'unistd.h': No such file or directory	libtcmalloc_minimal	D:\Users\LennoxHo\source\gperftools-main\src\common.cc	38		
Error	C1189	#error:  Could not determine cache line length - unknown architecture	page_heap_test	D:\Users\LennoxHo\source\gperftools-main\src\base\basictypes.h	379		
Error	C1189	#error:  Could not determine cache line length - unknown architecture	libtcmalloc_minimal	D:\Users\LennoxHo\source\gperftools-main\src\base\basictypes.h	379		
Error	C1189	#error:  Could not determine cache line length - unknown architecture	page_heap_test	D:\Users\LennoxHo\source\gperftools-main\src\base\basictypes.h	379		
Error	C1189	#error:  Could not determine cache line length - unknown architecture	libtcmalloc_minimal	D:\Users\LennoxHo\source\gperftools-main\src\base\basictypes.h	379		
Error	C1083	Cannot open include file: 'unistd.h': No such file or directory	page_heap_test	D:\Users\LennoxHo\source\gperftools-main\src\symbolize.cc	41		
Error	C1083	Cannot open include file: 'unistd.h': No such file or directory	libtcmalloc_minimal	D:\Users\LennoxHo\source\gperftools-main\src\symbolize.cc	41		
Error	C1189	#error:  Could not determine cache line length - unknown architecture	page_heap_test	D:\Users\LennoxHo\source\gperftools-main\src\base\basictypes.h	379		
Error	C1083	Cannot open include file: 'sys/resource.h': No such file or directory	thread_dealloc_unittest	D:\Users\LennoxHo\source\gperftools-main\src\tests\testutil.cc	41		
Error	C1083	Cannot open include file: 'unistd.h': No such file or directory	thread_dealloc_unittest	D:\Users\LennoxHo\source\gperftools-main\src\base\logging.h	44		
Error	C1083	Cannot open include file: 'unistd.h': No such file or directory	realloc_unittest	D:\Users\LennoxHo\source\gperftools-main\src\base\logging.h	44		
Error	C1083	Cannot open include file: 'sys/time.h': No such file or directory	frag_unittest	D:\Users\LennoxHo\source\gperftools-main\src\tests\frag_unittest.cc	40		
```
Deleting the stale `src/config.h` fixes those errors.

alk: We shouldn't bother about such corner cases. There is always git clean or similar. And both autotools and cmake support separate build directory builds.

For example, did you know that make distclean removes src/windows/gperftools/tcmalloc.h ? Yeah, one artifact we at some point might consider doing something about.

alk: And to re-iterate, IMHO this commit's proposed hardcoding of compiler flags knowledge (/FI and -include) is worse than any risks caused by improbable mixups you're referring to.

lennoxho: Yeah, you are probably right.

I have one last idea - would you be open to move the automake generated config.h from `<root>/src/` to another directory? say, `<root>/am/`? This can be another solution to avoiding the autotools/vsproj issue.

alk: But I am not sure I am ready yet to have src/windows/gperftools/tcmalloc.h generation to be amputated from autoconf (but on the other hand, already each time I do new release I already have to update more than 1 file manually; so I can be convinced perhaps with right arguments (but they have to be perfect))

Lets keep things as is for now. autotools does support separate directory builds and you're probably only person in the world with any risk of autotools vs msvc mixups.

alk: BTW I just looked at the diff between windows and non-windows tcmalloc.h. It differs in aligned new detection, which I am wondering we may consider doing via defined(__cpp_aligned_new) and by struct mallinfo, which availability we may consider detecting via templates magic (e.g. similar to recent getpc.h refactoring).

lennoxho: I can take a look at this in the follow up PR along with the HAVE_PTHREAD cleanups

alk:
Thanks a lot. Notably, latest revision failed couple tests. Needs this patch (yes, we should reconsider this TracePrintf business):

```
diff --git a/src/debugallocation.cc b/src/debugallocation.cc
index 5b6c19cb..a4d17cfb 100644
--- a/src/debugallocation.cc
+++ b/src/debugallocation.cc
@@ -911,6 +911,10 @@ static void TracePrintf(int fd, const char *fmt, ...) {
         l = va_arg(ap, size_t);
         base = 10;
         p++;
+      } else if (*p == 'z' && p[1] == 'x') {        // %zx
+        l = va_arg(ap, size_t);
+        base = 16;
+        p++;
       } else if (*p == 'u') {                       // %u
         l = va_arg(ap, unsigned int);
         base = 10;
```

I think this is minor enough for me to amend myself (I generally prefer not to touch contributed commits other than rebasing, but minor amends I do sometimes). Going to test and review some more and merge. Thanks a lot, again!

alk:
BTW, quick question. From windows user perspective, if we drop MSBuild support and have windows/msvc users use cmake, how convenient it is going to be for them ? (I think some test runs are less convenient via Visual Studio cmake integration, but some might even prefer it; plus we have greater unification)

I.e. now that MS appears to be officially supporting and even embracing cmake both in VS and vscode.

alk:
Another quick thing. So I just tried building in MSVC keeping project at VS2015 configuration. And it fails to build. Is it possible you forgot something minor?

One of complains is std::min crap in generic_writer.cc (I'll keep looking further)...

lennoxho:
> BTW, quick question. From windows user perspective, if we drop MSBuild support and have windows/msvc users use cmake, how convenient it is going to be for them

I'll personally be happy with the change, but I suspect it will be an unpopular decision among some developers.
Mostly because it will be yet another tool to pick up.

> One of complains is std::min crap in generic_writer.cc

Let me install the 140 build tools & try it out.

lennoxho:
I'm able to workaround the VS 2015 build issue 

```
--- a/src/windows/port.h
+++ b/src/windows/port.h
@@ -84,6 +84,7 @@
 #include <sys/types.h>       /* for _off_t */
 #include <assert.h>
 #include <stdlib.h>          /* for rand, srand, _strtoxxx */
+#include <algorithm>

 #if defined(_MSC_VER) && _MSC_VER >= 1900
 #define _TIMESPEC_DEFINED
```

Not entirely sure what changed.
Can you try this out? Thanks.

alk:
it was indeed lack of <algorithm> perhaps I missed this originally when doing generic_writer. Perhaps I only tested in later version of MSVC and got std::min by luck.

Anyways. Fixed now. And merged. Thanks a lot again.

Issue #1490, opened 2024-01-25
TCMalloc crashes on Windows 11 when opening a file dialog


lennoxho:
Here's a crash that consistently reproduces when the following conditions are met
1. The host machine is running Windows 11 with a build number between `[10.0.22000.0, 10.0.22621.900)`
2. The user has OneDrive Desktop syncing enabled
3. The application which is TCMalloc enabled tries to open a file dialog

Here's a [full breakdown](https://github.com/oneapi-src/oneTBB/issues/665) by TBBMalloc contributors.
Separate discussion for [mimalloc](https://github.com/microsoft/mimalloc/issues/498)

The TLDR is there is a `malloc`/`HeapFree` mismatch in the Windows explorer DLL.
This issue affects TCMalloc because while we replace calls to `malloc`, for `HeapFree` we just do a pass-through for instrumentation.

Here's a workaround that should work for TCMalloc https://github.com/lennoxho/gperftools/commit/18171db63b79f94ec15511109c9e813fe2551702

While MSFT has since issued a fix, Windows 11 `10.0.22000` aka 21H2 won't be EOL until October 2024.

With that in mind, do you think this is still worth addressing in TCMalloc?

lennoxho:
Added a brief comment and referenced this issue.
Created PR https://github.com/gperftools/gperftools/pull/1491

PR #1491, opened 2024-01-25
Work around malloc/HeapFree mismatch in early versions of Win11's shell32.dll

lennoxho:
This PR aims to work around a Windows 11 bug described in #1490

TLDR: Early versions of Windows 11 has a malloc/HeapFree mismatch which triggers when we try to open a file dialog.

**Testing done:**

Autotools:

- MSYS2-MinGW [link](https://github.com/msys2/msys2-installer/releases/tag/2024-01-13) + GCC 13.2

 
VSProj:
- MSVC 19.38 + WinSDK 10.0.22621.0

CMake:
- MSVC 19.38 + WinSDK 10.0.22621.0


alk:
Thanks!

PR #1498, opened 2024-03-26
ISO C++17 does not allow 'register' storage class specifier

hmihaylov-sc:
Variables in the `heap-checker_unittest` use the `register` storage class. This breaks builds that target C++17 and up.

### DIFF
--- a/src/tests/heap-checker_unittest.cc
+++ b/src/tests/heap-checker_unittest.cc
@@ -853,7 +853,7 @@ static void* HeapBusyThreadBody(void* a) {
 #elif defined(__x86_64__) && defined(__GNUC__)
   register int** ptr asm("r15");
 #else
-  register int** ptr;
+  int** ptr;
 #endif
   ptr = NULL;
   typedef std::set<int> Set;


alk:
Hi and thanks for the patch. When you say breaks, exactly what kind of breaks it is? The code already builds with -std=c++17 and is fine.

Your patch actually "breaks" the code's intention. Making asm declarations ineffective. I just checked (with pre-release of gcc 14 as shipped in Debian Sid). And yes it is already unportable and highly gcc- and platform-specific.

So what is it that you're fixing there? Can we do it in a way that doesn't break code's original intention (whatever questionable it might be) ?

alk:
On the other hand, clang already ignores asm+register thingy. With or without register keyword. But gcc still does (even upcoming release).

alk:
Oh, I see. It actually breaks the non-x86 builds... Odd. I thought I did non-x86 build tests recently.

alk:
Ah, ok. Clang and non-x86 (i.e. when we do non-asm thingy) fails to compile. So feel free to propose a change that removes "naked" register in the "#else" case. Rest we need to keep (for some definition of "need").

hmihaylov-sc:
Yes, sorry for the vague description. I'm building for `arm64` with `-Wall -Wextra -Wpendatic -Werror` with clang 17. The variable emit a warning, due to [register being deprecated in C++17](https://en.cppreference.com/w/cpp/language/storage_duration) and this in turn creates an error and compilation fails. I don't know why there is a difference in how this deprecated storage class is handled, because my x86 builds pass fine, not even a warning. I assume it is something specific for non-amd64 clang builds. I will add your proposed changes and leave the rest untouched.

Does it make more sense to have this case behind an `ifdef` for clang + non-x86 ?

alk:
Yes, removing "naked" register in the else case is what I want. Can you squash ?

To re-iterate, register + asm is non-standard, but sensible. clang still chooses to ignore (without warning or error), but gcc behaves. Naked register in the else case never made sense and is right to drop.

And also let me comment on -Werror. While in this case clang fails to compile naked register even without -Werror, the fact that you use it, is on your conscience. Compilers are known to produce bogus warnings from time to time. Then each minor update of the compiler introduces risk of breaking your builds. It is up to your of course. And we'll keep accepting patches that fix compiler warnings. It just irks me when people call fix of compiler warning as "fixing the build". (again, this case is proper fixing the build since clang fails even without -Werror).

alk:
Applied. Thanks.

hmihaylov-sc:
> Can you squash ?

Squashed :+1: 

> And also let me comment on -Werror. While in this case clang fails to compile naked register even without -Werror, the fact that you use it, is on your conscience. Compilers are known to produce bogus warnings from time to time. Then each minor update of the compiler introduces risk of breaking your builds. It is up to you of course. And we'll keep accepting patches that fix compiler warnings. It just irks me when people call fix of compiler warning as "fixing the build". (again, this case is proper fixing the build since clang fails even without -Werror).

I agree, it's not correct to say that a warning `breaks the build`, its only hints at a potential issue. I'll change the comment above. I run with `-Werror` because the warnings I've enabled rarely if ever produce false positives. Thanks for accepting the patch.

Issue #1501, opened 2024-04-10
remove this line, mul define


zhangdexin:
https://github.com/gperftools/gperftools/blob/7c0cbb3a0c4a79c267eeb59deb2c657e91ae8ad1/src/base/proc_maps_iterator.cc#L363

alk:
Thanks for spotting this. Your patch is now applied.

PR #1502, opened 2024-04-10
Update proc_maps_iterator.cc remove flags_tmp for multi define

zhangdexin:
https://github.com/gperftools/gperftools/issues/1501


alk:
Applied. thanks. (I updated commit message)

PR #1519, opened 2024-05-24
generic_fp stacktrace: aarch64 frame pointer may be 8 byte aligned

blahgeek:
The aarch64 ABI does not specify the alignment of the frame pointer, it only specify the alignment of the stack frame.

From [the ABI doc](https://github.com/ARM-software/abi-aa/blob/main/aapcs64/aapcs64.rst#the-frame-pointer):

> SP mod 16 = 0. The stack must be quad-word aligned.
> ...
> The location of the frame record within a stack frame is not specified.

Clang in particularly would generate such code. [Example](https://godbolt.org/z/nan8crMGf):


alk:
Thanks a lot. Applied.

PR #1520, opened 2024-05-31
win32: don't patch on arm64 machine

Chilledheart:
arm64's patch is incomplete and conflicts with system's native jit.

see https://learn.microsoft.com/en-us/windows/win32/winprog64/wow64-implementation-details

Short story: gperftools' win32 patcher doesn't get prepared for arm64 assembly. It could fail in silence if the patch conflicts with the system's x86 to arm64 jit enginee.

### DIFF
--- a/src/windows/patch_functions.cc
+++ b/src/windows/patch_functions.cc
@@ -997,6 +997,43 @@ BOOL WINAPI WindowsInfo::Perftools_FreeLibrary(HMODULE hLibModule) {
   return rv;
 }
 
+// see https://learn.microsoft.com/en-us/answers/questions/518344/tell-if-process-archtitecture-is-arm64-(x64-compat
+typedef BOOL (WINAPI* PFNISWOW64PROCESS2)(HANDLE hProcess,
+                                          USHORT *pProcessMachine,
+                                          USHORT *pNativeMachine);
+#ifndef IMAGE_FILE_MACHINE_UNKNOWN
+#define IMAGE_FILE_MACHINE_UNKNOWN 0x0
+#endif // IMAGE_FILE_MACHINE_UNKNOWN
+#ifndef IMAGE_FILE_MACHINE_ARM64
+#define IMAGE_FILE_MACHINE_ARM64 0xAA64
+#endif // IMAGE_FILE_MACHINE_ARM64
+
+static BOOL IsWow64Process2Wrapper(HANDLE hProcess,
+                                   USHORT *pProcessMachine,
+                                   USHORT *pNativeMachine) {
+  HMODULE m = GetModuleHandleW(L"Kernel32.dll");
+  assert(m && "kernel32 dll not loaded");
+  static const auto fPointer =
+    reinterpret_cast<PFNISWOW64PROCESS2>(
+      reinterpret_cast<void*>(::GetProcAddress(m, "IsWow64Process2")));
+  if (fPointer == nullptr) {
+    ::SetLastError(ERROR_CALL_NOT_IMPLEMENTED);
+    return FALSE;
+  }
+  return fPointer(hProcess, pProcessMachine, pNativeMachine);
+}
+
+static bool IsArm64NativeMachine() {
+  USHORT ProcessMachine = IMAGE_FILE_MACHINE_UNKNOWN;
+  USHORT NativeMachine = IMAGE_FILE_MACHINE_UNKNOWN;
+
+  if (!IsWow64Process2Wrapper(GetCurrentProcess(),
+                              &ProcessMachine, &NativeMachine)) {
+    return false;
+  }
+  return NativeMachine == IMAGE_FILE_MACHINE_ARM64;
+}
+
 
 // ---------------------------------------------------------------------
 // PatchWindowsFunctions()
@@ -1011,7 +1048,9 @@ void PatchWindowsFunctions() {
                            disable_env[0] == '1' &&
                            disable_env[1] == '\0';
 
-  if (!should_skip) {
+  const bool is_arm64_native = IsArm64NativeMachine();
+
+  if (!should_skip && !is_arm64_native) {
     // This does the libc patching in every module, and the main executable.
     PatchAllModules();
     main_executable_windows.Patch();


Chilledheart:
The crash without this patch could be repeated via calling `SHGetFolderPathW` from a x86/x64 tcmalloc-linked binary on a windows 11 arm64 machine.

alk:
Thanks for the patch, but I'll need a lot more details what is being fixed here. Are you saying MS somehow fails to emulate x64 correctly or what ?

Chilledheart:
Correct. If you run executable with tcmalloc linked (both of x86 and x64, both mingw and msvc) on an arm64 windows operating system, you are likely to get killed. With this patch, it is safe to run.

If you cannot repeat somehow, please try building an executable containing above call or building a gui executable.

Chilledheart:
> Thanks for the patch, but I'll need a lot more details what is being fixed here. Are you saying MS somehow fails to emulate x64 correctly or what ?

In more precise terms, x86 is emulated but x64 is jitted. Both cases fail with the same reason.

You can refer to the link in commit message or just get the list of loaded DLL libraries of an x64 process on an arm64 windows. There are many informations on the internet as well.

Chilledheart:
It is a simple patch. The code explains it all. If you need more comments, I would say it simply skips patching which is causing troubles on an arm64 windows. IsWowProcess2 is introduced in windows 10 and it is guaranteed to set native machine property to arm64 if and only if the executable runs on an arm64 windows.

Chilledheart:
This is one way of the only two ways to tell whether it is arm64 windows. The other way is checking environment variables (pretty dirty, I thought). Please refer to the link in message 0.

alk:
I am sorry, but I don't get why we have to patch workaround for what you claim to be a failure of MS to keep their promise of having functional x64 support on arm64 machines? Why not report this as a bug to them ?

Chilledheart:
I don't attach the binary to a debugger such as windbg or something like that. But I only get noticed via Event Viewer. If you want to rootcause the failure and improve the patch, please do it. I only guarantee this patch works.

Chilledheart:
> I am sorry, but I don't get why we have to patch workaround for what you claim to be a failure of MS to keep their promise of having functional x64 support on arm64 machines? Why not report this as a bug to them ?

For short, I don't know if jit program is guaranteed to run via emulation just like chrome. And I won't work for Microsoft. There is no software department of Microsoft in my country.

alk:
And you need "software department of Microsoft in my country" to simple report bugs? ;)

To look at this from another perspective, it is entirely possible that our patching code has bugs (plural). But simply saying, "well, on version/platform it doesn't work so lets turn it off", is not how we fix bugs.

Chilledheart:
In my understanding, you cannot jit x64 assemblies into an arm64 process because the cpu won't execute x64 intrinsics. For example, if you run v8 engine on arm64 inside an x64 chrome executable/process, you might get troubles without patch.

Chilledheart:
> And you need "software department of Microsoft in my country" to simple report bugs? ;)
> 
> 
> 
> To look at this from another perspective, it is entirely possible that our patching code has bugs (plural). But simply saying, "well, on version/platform it doesn't work so lets turn it off", is not how we fix bugs.

Yes. The users team won't answer or report to the software team.

For bugs, I would say tcmalloc works for x86/x64 windows and no obvious bugs are found. It is only a nice feature to run for arm64 windows directly instead of switching to link with arm64 tcmalloc.

alk:
I cannot take this patch without understanding what is the thing we're fixing or working around.

Perhaps, consider explaining more carefully and more in-depth what is it, that you're fixing. And also please explain why not patching (which effectively disables libtcmalloc) is the fix rather than simply not linking.

Chilledheart:
Sure. Let me explain again.
For tcmalloc internals, it patches x86 on x86 process via PatchWindowsFunctions and patches x64 on x64 process via the same function. This function will patch every dll's HeapAlloc and HeapFree via assemblies and patch LoadLibray to patch the library's calls of HeapAlloc and HeapFree via assemblies. tcmalloc doesn't patch any arm64 process for now.
Screenshot 2024-06-01 at 10 08 03

For windows internals, the above way works for x86 even on x64 operating system because the execuable runs on x86 emulation (via wow64.dll).
Screenshot 2024-06-01 at 10 09 54

For arm64 operating system, it becomes different because only x86 runs on emulation not x64 (via xtajit.dll).

Screenshot 2024-06-01 at 10 04 37
Screenshot 2024-06-01 at 10 13 00

(x86 via xtajit.dll ProcessMachine: IMAGE_FILE_MACHINE_I386)

Screenshot 2024-06-01 at 10 05 15
(x64 via xtajit64.dll ProcessMachine: IMAGE_FILE_MACHINE_UNKNOWN)
Screenshot 2024-06-01 at 10 13 14

I believe the failing case happens on x64 binaries running arm64 because the MS way is no longer emulation but JIT. If you patch x64 assemblies on an native arm64 (JITted via xtajit64.dll) process, the cpu won't run any x64 intrinsics.

Screenshot 2024-06-01 at 09 57 40

| ProcessMachine  |  x86 os | x64 os  | arm64os  | 
|---|---|---|---|
|  x86 binary |  IMAGE_FILE_MACHINE_I386 | IMAGE_FILE_MACHINE_I386  |  IMAGE_FILE_MACHINE_I386 | 
|  x64 binary | x |  IMAGE_FILE_MACHINE_AMD64 |  IMAGE_FILE_MACHINE_UNKNOWN | 
|  arm64 binary | x |  x |  IMAGE_FILE_MACHINE_ARM64 | 

IMAGE_FILE_MACHINE_AMD64/IMAGE_FILE_MACHINE_I386/IMAGE_FILE_MACHINE_ARM64: Native/Emulation
IMAGE_FILE_MACHINE_UNKNOWN: jitted

Chilledheart:
Hopefully, you might know more internals of tcmalloc than me. If you need more explains, please let me know.

Chilledheart:
cc-ed windows team via X. 

image

Chilledheart:
And the patch would be improved by detecting if process is JITed (IMAGE_FILE_MACHINE_UNKNOWN) than detecting arm64 (IMAGE_FILE_MACHINE_ARM64) so we can exclude x86 binaries from disabling tcmalloc (because it is never JITed).

Chilledheart:
For record, I tried windbg (both of x64 and arm64 versions), the problem doesn't appear. I am looking for something that generate crash report for windows applications.

Chilledheart:
For record, if we use chromium's [delayload trick](https://source.chromium.org/chromium/chromium/src/+/main:build/config/win/BUILD.gn;l=382) with ucrt (not msvcrt), it is possible to run x64 program on top of arm64 windows without disabling tcmalloc.

image

This trick doesn't apply to gcc (mingw) because only lld linker supports msvc-like `--delayload=dll_name` option.

Chilledheart:
> In my understanding, you cannot jit x64 assemblies into an arm64 process because the cpu won't execute x64 intrinsics. For example, if you run v8 engine on arm64 inside an x64 chrome executable/process, you might get troubles without patch.

I think my previous guess is wrong. someone clarified the v8 (jit issue) at x.com.
Screenshot 2024-06-01 at 18 02 34

Considerring https://github.com/gperftools/gperftools/pull/1520#issuecomment-2143317194, there is something wrong with  actual patching. However I still cannot start debugging because I don't know how to create crash report from it.

alk:
I still don't understand you. It is particularly puzzling how you oppose jitting and emulating. Perhaps in windows world those words have somewhat special meaning (i.e. like "WORD" still being 16 bits :) ), but to me from general software engineering perspective jit is just one way to do emulation. I.e. implementation detail. Why that matters is unclear to me.

With that said, here is one guess what might be happening here. libtcmalloc doing "patching" of code that calls to heap routines is by design going into guts of all other loaded dlls. And it only understands very specific subset of instructions. So it is by design super-brittle. I can imagine that if (and that is "if" you could help clarify) we're dealing with processes that load ***both*** x64 and arm64 dlls. Then we could be in trouble. My understanding is MS went ahead and somehow choose to support this. And I am not sure if this is brave and noble or just stupid.

But at least this use-case (i.e. mixing arm64 dlls into x64 process) would clearly be not possible to support sensibly. And who knows maybe MS even choose to have "native" and thus faster dlls always loaded for stuff like system's dlls even for x64 processes. Then we're in more trouble.

Either way, I am not sure if simply disabling patching and those "breaking" tcmalloc is the way to go. Especially that you're claiming that it "works" in some cases.

BTW it might be useful for you to look at how mimalloc is handling this. They also offer that patching way and should have same or similar problems (via mimalloc-redirect which is built against some MS open-source general patching library which I struggling to locate now). And yes we're open to accepting commit(s) to reimplement our patching to use that MS more general and more expertful patching thingy. If you're open to that

Chilledheart:
> I still don't understand you. It is particularly puzzling how you oppose jitting and emulating. Perhaps in windows world those words have somewhat special meaning (i.e. like "WORD" still being 16 bits :) ), but to me from general software engineering perspective jit is just one way to do emulation. I.e. implementation detail. Why that matters is unclear to me.
> 
> 
> 
> With that said, here is one guess what might be happening here. libtcmalloc doing "patching" of code that calls to heap routines is by design going into guts of all other loaded dlls. And it only understands very specific subset of instructions. So it is by design super-brittle. I can imagine that if (and that is "if" you could help clarify) we're dealing with processes that load ***both*** x64 and arm64 dlls. Then we could be in trouble. My understanding is MS went ahead and somehow choose to support this. And I am not sure if this is brave and noble or just stupid.
> 
> 
> 
> But at least this use-case (i.e. mixing arm64 dlls into x64 process) would clearly be not possible to support sensibly. And who knows maybe MS even choose to have "native" and thus faster dlls always loaded for stuff like system's dlls even for x64 processes. Then we're in more trouble.
> 
> 
> 
> Either way, I am not sure if simply disabling patching and those "breaking" tcmalloc is the way to go. Especially that you're claiming that it "works" in some cases.
> 
> 
> 
> BTW it might be useful for you to look at how mimalloc is handling this. They also offer that patching way and should have same or similar problems (via mimalloc-redirect which is built against some MS open-source general patching library which I struggling to locate now). And yes we're open to accepting commit(s) to reimplement our patching to use that MS more general and more expertful patching thingy. If you're open to that

mimalloc doesn't work at all.

alk:
Well, thats actually somewhat of a good news IMO. Now you can bug them instead. They're part of MS and definitely more experful on all things windows. And despite lack of MS representation in your country ;) they're likely to respond. And once the fix and/or better understanding of the problem emerges you can come back here and offer a solution.

Chilledheart:
> Well, thats actually somewhat of a good news IMO. Now you can bug them instead. They're part of MS and definitely more experful on all things windows. And despite lack of MS representation in your country ;) they're likely to respond. And once the fix and/or better understanding of the problem emerges you can come back here and offer a solution.

At least, I have a working solution through it doesn't work for windows arm64 natively. But for gperftools, it is incomplete workaround because it only works for msvc compiler and clang(lld linker). It still needs troubleshooting to locate the bug in the dark side of hotpatching and x64 jit but sadly Microsoft doesn't provide a way to do it. windbg and logging doesn't help.

Chilledheart:
In short, it can be worked around but doesn't work for gcc compiler.

alk:
Well, we don't have arm64 support on windows. (mimalloc's patching is likely to support it, btw; but not necessary "hybrid" thingy). And with lack of arm64 support, we don't have to work around something we don't understand. IMHO.

With that, I am closing this pull request. I understand that from your perspective it looks like a fix and small adjustment of our code. But we have single maintainer (me) who is not super-qualified on things around windows. And I cannot accept patch that "just" fixes something I don't understand. Because I won't be able to test it or maintain this code.

If you have strong arguments, please, feel free to reopen.

Chilledheart:
With additional help of crash dump, it said the crash in arm64 windows is due to heap corruption.

Screenshot 2024-06-08 at 13 15 15

Chilledheart:
Screenshot 2024-06-08 at 13 15 54

PR #1539, opened 2024-07-26
Fixed an error in ppref

maxim-lbmv:
When searching for symbols in the analyzed program, if the program is analyzed on another machine, the paths to the program executable file specified in the pprof call argument and the path specified in the analysis results file may differ. Therefore, the condition of comparing these paths does not work. It is necessary to compare not the paths, but the names of executable files.

### DIFF
--- a/src/pprof
+++ b/src/pprof
@@ -70,6 +70,7 @@
 
 use strict;
 use warnings;
+use File::Basename;
 use Getopt::Long;
 use Cwd;
 use POSIX;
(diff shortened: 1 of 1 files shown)


alk:
Hi and thanks for the patch, but we're not updating our (perl) version of pprof anymore. Consider making sure that whatever you're fixing/improving is supported by github.com/google/pprof (golang version; much much improved)

maxim-lbmv:
But the ppref file is still in your repository, maybe you can make an exception and accept the patch? It's only two lines. I compile and debug my program on a PC and run it on a small device, so the paths of the executable file on the PC and on the device do not coincide and as a result of analysis the addresses of functions are not replaced by their names.

alk:
I understand, but I am sorry, I cannot verify your commit. You can always do your local patch(es) if you insist on using older perl pprof.

PR #1543, opened 2024-08-06
stacktrace_libunwind: Use unw_backtrace to compute backtrace faster

### LINKED ISSUE #1277: libunwind based backtracer appears to be much slower than the one from stacktrace_x86-inl.h
MongoDB currently vendors `gperftools-2.7` but we are working on upgrading to `gperftools-2.9`. We had arranged things, with some admitted fragility, such that `libunwind` was an option available to the logic to select a stack trace collection backend, but `TCMALLOC_DONT_PREFER_LIBUNWIND` ended up defined (basically, we defined `HAVE_LIBUNWIND_H` but did not define `NO_FRAME_POINTER`). That ensured that on platforms with customized unwinders, we wouldn't use libunwind, but we could still fall back to it if no arch specific unwinder was available.

We did that because we had found that for our DEBUG builds (which use `debugallocation.cc`) that they ran dramatically slower when using the `libunwind` based stack collection routines. So much so that they would time out in our CI system: what was normally a 5 minute task was instead timing out after 90 minutes.

For `gperftools-2.9` we will need to do something different to achieve the same effect because the special case handling which examined `NO_FRAME_POINTER` to decide whether to set `TCMALLOC_DONT_PREFER_LIBUNWIND` was removed for `x86_64` sometime after `gperftools-2.7` (and, please see #1276 for why we can't just set `TCMALLOC_DONT_PREFER_LIBUNWIND` ourselves, which would work nicely if it was possible). But, we will figure something out. Most likely, we will simply stop telling gperftools that we `HAVE_LIBUNWIND_H` if we are doing a `DEBUG` build on `x86_64`, which we know will be OK because there is a custom unwind for that platform, and we don't need the fallback to `libunwind` to be available.

However, long term we would like to standardize on libunwind. We have already done so for our own internal backtrace collection. But we cannot do so for our gperftools build given the current performance of `debugallocation` when using libunwind. I'm reporting this here first in case the issue is simply an unoptimized or erroneous use of the `libunwind` API in gperftools, and that the performance could be improved by making changes to how `gperftools` uses `libunwind`. If that isn't possible, then probably the next step is to take the matter over to the `libunwind` developers. Meanwhile, we will also be double checking our `libunwind` integration to make sure we haven't done something unfortunate that is undermining its performance.

If possible, please improve the performance of the libunwind based stack collector.

thecodingwizard:
Previously, stacktrace_libunwind used unw cursors to compute the backtrace, which is slow. This PR updates stacktrace_libunwind's `GetStackTrace()` function to compute the backtrace using `unw_backtrace`, which is substantially faster.

The main benefit of this change is that malloc profiling becomes substantially faster. I have seen > 5x profiling performance improvements with this change.

I was also able to reproduce @acmorrow's benchmark results from #1277 on my machine. The current libunwind's `GetStackTrace()` implementation performance is as follows:

```
--------------------------------------------------------------------------------
Benchmark                      Time             CPU   Iterations UserCounters...
--------------------------------------------------------------------------------
BM_GetStackTrace/1          3832 ns         3828 ns       182886 items_per_second=261.232k/s
BM_GetStackTrace/8          7705 ns         7692 ns        95400 items_per_second=130.008k/s
BM_GetStackTrace/64        31017 ns        31003 ns        22564 items_per_second=32.2546k/s
BM_GetStackTrace/512      184911 ns       184846 ns         3789 items_per_second=5.4099k/s
BM_GetStackTrace/4096    1356384 ns      1353508 ns          491 items_per_second=738.821/s
BM_GetStackTrace/8192    2578422 ns      2564196 ns          273 items_per_second=389.986/s
```

With this PR, the performance improves substantially:

```
--------------------------------------------------------------------------------
Benchmark                      Time             CPU   Iterations UserCounters...
--------------------------------------------------------------------------------
BM_GetStackTrace/1           102 ns          102 ns      6913378 items_per_second=9.84186M/s
BM_GetStackTrace/8           166 ns          166 ns      4218120 items_per_second=6.02384M/s
BM_GetStackTrace/64          517 ns          517 ns      1355825 items_per_second=1.93412M/s
BM_GetStackTrace/512        3252 ns         3251 ns       216073 items_per_second=307.559k/s
BM_GetStackTrace/4096      25256 ns        25194 ns        27830 items_per_second=39.6917k/s
BM_GetStackTrace/8192      50328 ns        50238 ns        13935 items_per_second=19.9051k/s
```

I have tested this change with `make check` and `make stacktrace_unittest && ./stacktrace_unittest`.

I would greatly appreciate any feedback regarding this PR. Thank you!

Closes #1277

### DIFF
--- a/src/stacktrace.cc
+++ b/src/stacktrace.cc
@@ -277,13 +277,15 @@ static GetStackImplementation *get_stack_impl;
 #if 0
 // This is for the benefit of code analysis tools that may have
 // trouble with the computed #include above.
-# include "stacktrace_libunwind-inl.h"
 # include "stacktrace_generic-inl.h"
+# include "stacktrace_libgcc-inl.h"
+# include "stacktrace_libunwind-inl.h"
 # include "stacktrace_generic_fp-inl.h"
 # include "stacktrace_powerpc-linux-inl.h"
-# include "stacktrace_win32-inl.h"
+# include "stacktrace_powerpc-darwin-inl.h"
 # include "stacktrace_arm-inl.h"
 # include "stacktrace_instrument-inl.h"
+# include "stacktrace_win32-inl.h"
 #endif
 
 static void init_default_stack_impl_inner(void);
--- a/src/stacktrace_libunwind-inl.h
+++ b/src/stacktrace_libunwind-inl.h
@@ -81,18 +81,45 @@ static __thread int recursive ATTR_INITIAL_EXEC;
 //   int skip_count: how many stack pointers to skip before storing in result
 //   void* ucp: a ucontext_t* (GetStack{Trace,Frames}WithContext only)
 static int GET_STACK_TRACE_OR_FRAMES {
-  void *ip;
   int n = 0;
+  if (recursive) {
+    return 0;
+  }
+  ++recursive;
+
+#if (!IS_WITH_CONTEXT && !IS_STACK_FRAMES)
+  // GetStackTrace(): Use unw_backtrace() to get the backtrace more quickly.
+
+  // From malloc_hook.cc, the maximum backtrace size that's requested is
+  // <= 32 + 6 + 3 + 1 + max_depth. We add another 2 for GetStackTrace()'s
+  // skip_count.
+  constexpr size_t kMaxBacktraceSize = 32 + 6 + 3 + 1 + 32 + 2;
+  static __thread void *libunwind_backtrace_result[kMaxBacktraceSize];
+
+  skip_count += 2;
+  if (PREDICT_FALSE(skip_count + max_depth > kMaxBacktraceSize)) {
+    fprintf(stderr,
+            "error: exceeded max backtrace size of %lu. max depth = %d, skip "
+            "count = %d.\n",
+            kMaxBacktraceSize, max_depth, skip_count);
+    exit(1);
+  }
+  n = unw_backtrace(libunwind_backtrace_result, max_depth + skip_count);
+  n -= skip_count;
+  if (PREDICT_FALSE(n <= 0)) {
+    n = 0;
+    goto out;
+  }
+
+  std::copy(libunwind_backtrace_result + skip_count,
+            libunwind_backtrace_result + skip_count + n, result);
+#else
+  void *ip;
   unw_cursor_t cursor;
   unw_context_t uc;
 #if IS_STACK_FRAMES
   unw_word_t sp = 0, next_sp = 0;
 #endif
-
-  if (recursive) {
-    return 0;
-  }
-  ++recursive;
 
 #if (IS_WITH_CONTEXT && defined(BASE_STACKTRACE_UNW_CONTEXT_IS_UCONTEXT))
   if (ucp) {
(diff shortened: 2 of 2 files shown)


src/stacktrace.cc
```diff
@@ -277,13 +277,15 @@ static GetStackImplementation *get_stack_impl;
...
 # include "stacktrace_generic-inl.h"
+# include "stacktrace_libgcc-inl.h"
+# include "stacktrace_libunwind-inl.h"
 # include "stacktrace_generic_fp-inl.h"
 # include "stacktrace_powerpc-linux-inl.h"
-# include "stacktrace_win32-inl.h"
+# include "stacktrace_powerpc-darwin-inl.h"
 # include "stacktrace_arm-inl.h"
 # include "stacktrace_instrument-inl.h"
+# include "stacktrace_win32-inl.h"
```
thecodingwizard, line 288: This reorders the include statements to match the order in which they are included. It also adds a missing `libgcc-inl.h` include statement.

alk:
Thanks a lot for the patch.

But the code appears to be still doing unbounded caching within libunwind. In practice it won't grow much, usually, but who knows. I.e.: https://github.com/libunwind/libunwind/blob/fb917e18e8bef992ddd4179143a65a59f18f00ca/src/x86_64/Gtrace.c#L274

So perhaps we make it optional? Also I see unw_backtrace2 which can unwind from unw_context.

As for the patch; the max backtrace length thingy and fprintf with exit() is not great. We have places where we do quite a bit of backtracing depth, more than 32. Also you sure you want to do a static per-thread temporary backtraces buffer?

thecodingwizard:
Thank you for taking a look!

> So perhaps we make it optional?

What's the best way to do this? Maybe make a `stacktrace_libunwind-fast-inl.h` file and add `libunwind_fast` as a new stacktrace option? (For the other three functions that aren't `GetStackTrace()`, we can call the libunwind versions of those functions to avoid code duplication.)

> Also I see unw_backtrace2 which can unwind from unw_context.

Perhaps I'll add a comment for now and leave that for a second PR? Please let me know if you'd rather I do both at once though.

> As for the patch; the max backtrace length thingy and fprintf with exit() is not great.

I agree. Do you know of a better way to do this? I wasn't sure whether using `malloc()` inside this function would be safe (and even if it were, I was worried that would be too slow). However, I'm not sure how to properly handle `skip_count` without using a temporary array.

Thanks again for your feedback!

alk:
Yes, splitting this into new distinct stacktracing method is fine.

As for how to deal with the storage, how about doing it on stack. Say 256 or 128 entries. Then whatever doesn't fit, gets truncated (no big deal). Another easy win here is if you're given more storage than whatever you have on stack, use it rather than your storage.

alk:
I apologize for loosing track of this.

So looking at it just now I have few comments. But main one is libunwind by default already aliases backtrace to unw_backtrace. So then why not use some of the tricks you did here to "unbreak" stacktrace_generic instead ? (It was for long time hardcoded but since 2019 libunwind has configure flag to disable it: https://github.com/libunwind/libunwind/pull/142)

PR #1544, opened 2024-08-07
fix broken qnx support

iDings:
Link regex libray explicitly, and skip header when parse proc maps. Disable libgcc unwinder because it always have only one stackback record.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -34,6 +34,8 @@ if(MINGW OR MSVC)
   set(DEFAULT_BUILD_DEBUGALLOC OFF)
 elseif(CYGWIN)
   set(DEFAULT_BUILD_CPU_PROFILER OFF)
+elseif(CMAKE_SYSTEM_NAME MATCHES QNX)
+  set(DEFAULT_BUILD_CPU_PROFILER OFF)
 endif()
 
 # Heap checker is Linux-only (and deprecated).
@@ -466,6 +468,10 @@ link_libraries(psapi synchronization shlwapi)
 
 endif()
 
+if(CMAKE_SYSTEM_NAME MATCHES QNX)
+  link_libraries(socket)
+endif()
+
 if(BUILD_TESTING)
   add_library(gtest STATIC vendor/googletest/googletest/src/gtest_main.cc
     vendor/googletest/googletest/src/gtest-assertion-result.cc
@@ -483,6 +489,10 @@ if(BUILD_TESTING)
 
   target_include_directories(gtest INTERFACE
     $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/vendor/googletest/googletest/include>)
+
+  if(CMAKE_SYSTEM_NAME MATCHES QNX)
+    target_link_libraries(gtest INTERFACE regex)
+  endif()
 
   add_executable(low_level_alloc_unittest src/base/low_level_alloc.cc
           src/malloc_hook.cc
@@ -810,7 +820,7 @@ if(GPERFTOOLS_BUILD_HEAP_CHECKER OR GPERFTOOLS_BUILD_HEAP_PROFILER)
     set(sampling_test_SOURCES src/tests/sampling_test.cc)
     add_executable(sampling_test src/tests/sampling_test.cc)
     target_compile_definitions(sampling_test PRIVATE PPROF_PATH=${CMAKE_CURRENT_SOURCE_DIR}/src/pprof)
-    target_link_libraries(sampling_test tcmalloc)
+    target_link_libraries(sampling_test tcmalloc $<$<PLATFORM_ID:QNX>:regex>)
     add_test(sampling_test sampling_test)
 
     if(GPERFTOOLS_BUILD_HEAP_PROFILER)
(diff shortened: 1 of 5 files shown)


alk:
Thanks for the patch. Looks good.

Is it hard for you to split off proc-maps parsing change? (I'd like, if possible, to merge it separately)

As for the rest:

* I am not super enthusiastic about adding -lregex which is only used by tests to "global" linker flags. I guess, the alternative is more complexity in the cmake/autotools bits so, okay, let have it.
* Can you leave _Unwind_Backtrace thingy in qnx but ensure it is not picked as default instead ? There is chance qnx will eventually behave and make it right, since this is same tech as used by exceptions and it should work.

iDings:
Thanks for your review.

>Is it hard for you to split off proc-maps parsing change? (I'd like, if possible, to merge it separately)
 
I  will split proc-maps parsing change

> I am not super enthusiastic about adding -lregex which is only used by tests to "global" linker flags. I guess, the alternative is more complexity in the cmake/autotools bits so, okay, let have it.

yeah, this one is not so good, for CMake, one method is to add -lregex to each tests if need such as
```
-  target_link_libraries(low_level_alloc_unittest common gtest)
+  target_link_libraries(low_level_alloc_unittest common gtest $<$<PLATFORM_ID:QNX>:regex>)
```

And because the library is used by gtest, maybe gtest CMake or autotools should export the needed link library flag.

> Can you leave _Unwind_Backtrace thingy in qnx but ensure it is not picked as default instead ? There is chance qnx will eventually behave and make it right, since this is same tech as used by exceptions and it should work.

I will leave it,  from testing for now, _Unwind_Backtrace seems not work correctly, there are always only one stacktrace record.

iDings:
@alk I had uploaded the modified commits which split off the one big commit to three irreducible small commits.
Pls review these commits when you get a moment, thanks.

alk:
Thanks a lot. This is great. And especially, huge thanks for splitting this up.

So there is only one "gotcha" around autotools part of the regex change. See here: https://www.gnu.org/software/automake/manual/html_node/Program-Variables.html

Basically if you override some per-target flag you typically need to manually add AM_LDFLAGS. But also please check if this override is necessary at all. I'd hope that libtool might already automagically add -lregex when linking with gtest.la (.la records libraries needed to link with).

So I took 2 of 3 commits and skipped regexp bit for now.

iDings:
Thanks for review the patches. 
Yeah, the regex part is a little dirty,  need to find a better  solution.

iDings:
Thanks for reviewing the pathes.
Yeah,  the regex patch is a little dirty, need another better solution.

alk:
Hm, so I thought that I've shared my variant of fixing autotools part of regex issue but apparently I haven't.

Here is what I have (and intend to merge then; originally, I was sure I passed it over to you encouraging you to integrate with your change, but looks like it was somehow lost)

```
commit 0e013e50dc53cb61b515c0b61828aae6d402772a (wip-qnx-regex)
Author: Aliaksey Kandratsenka <alkondratenko@gmail.com>
Date:   Sun Aug 18 15:28:56 2024 -0400

    wip: qnx regex thingy

diff --git a/Makefile.am b/Makefile.am
index 4fa6349d..1870eeda 100644
--- a/Makefile.am
+++ b/Makefile.am
@@ -163,6 +163,7 @@ libgtest_la_SOURCES = vendor/googletest/googletest/src/gtest_main.cc \
 
 libgtest_la_CPPFLAGS = -I$(top_srcdir)/vendor/googletest/googletest/include \
  -I$(top_srcdir)/vendor/googletest/googletest/ $(AM_CPPFLAGS)
+libgtest_la_LIBADD = $(REGEX_LIBS)
 
 gtest_CPPFLAGS = -I$(top_srcdir)/vendor/googletest/googletest/include $(AM_CPPFLAGS)
 
@@ -716,7 +717,7 @@ TESTS += sampling_test
 sampling_test_SOURCES = src/tests/sampling_test.cc
 sampling_test_LDFLAGS = $(TCMALLOC_FLAGS) $(AM_LDFLAGS)
 sampling_test_CPPFLAGS = $(AM_CPPFLAGS) "-DPPROF_PATH=$(top_srcdir)/src/pprof"
-sampling_test_LDADD = libtcmalloc.la
+sampling_test_LDADD = libtcmalloc.la $(REGEX_LIBS)
 
 endif WITH_HEAP_PROFILER_OR_CHECKER
 
@@ -790,7 +791,7 @@ sampling_debug_test_SOURCES = $(sampling_test_SOURCES)
 sampling_debug_test_CXXFLAGS = $(sampling_test_CXXFLAGS)
 sampling_debug_test_CPPFLAGS = $(sampling_test_CPPFLAGS)
 sampling_debug_test_LDFLAGS = $(sampling_test_LDFLAGS)
-sampling_debug_test_LDADD = libtcmalloc_debug.la
+sampling_debug_test_LDADD = libtcmalloc_debug.la $(REGEX_LIBS)
 
 endif WITH_HEAP_PROFILER_OR_CHECKER
 
diff --git a/configure.ac b/configure.ac
index a2b8588b..6eb2dab6 100644
--- a/configure.ac
+++ b/configure.ac
@@ -276,6 +276,18 @@ AC_CHECK_HEADERS(asm/ptrace.h)           # get ptrace macros, e.g. PT_NIP
 # check for socketpair, some system, such as QNX, need link in an socket library to use it
 AC_SEARCH_LIBS([socketpair], [socket])
 
+REGEX_LIBS=
+# "sufficiently unix" systems need regexec for unit tests
+if test "x$need_nanosleep" = xyes; then
+  save_LIBS="$LIBS"
+  LIBS="$REGEX_LIBS"
+  # QNX needs -lregex; but lets test just in case
+  AC_SEARCH_LIBS([regexec], [regex], [], [AC_MSG_ERROR([failed to locate regexec() (used for tests)])])
+  REGEX_LIBS="$LIBS"
+  LIBS="$save_LIBS"
+fi
+AC_SUBST(REGEX_LIBS)
+
 # We override a lot of memory allocation routines, not all of which are
 # standard.  For those the system doesn't declare, we'll declare ourselves.
 AC_CHECK_DECLS([cfree,
```

iDings:
I loss some time to complete some other things.thanks for fix it.

PR #1551, opened 2024-08-29
fix compiler error when NO_TCMALLOC_SAMPLES is defined

lano1106:
the error is:
src/malloc_backtrace.cc:51:5: error: redefinition of 'int tcmalloc::GrabBacktrace(void**, int, int)'
   51 | int GrabBacktrace(void** result, int max_depth, int skip_count) {
      |     ^~~~~~~~~~~~~
In file included from src/malloc_backtrace.cc:34:
src/malloc_backtrace.h:43:12: note: 'int tcmalloc::GrabBacktrace(void**, int, int)' previously defined here
   43 | inline int GrabBacktrace(void** result, int max_depth, int skip_count) { return 0; }

### DIFF
--- a/src/malloc_backtrace.cc
+++ b/src/malloc_backtrace.cc
@@ -41,7 +41,7 @@ extern "C" {
 }
 
 namespace tcmalloc {
-
+#ifndef NO_TCMALLOC_SAMPLES
 // GrabBacktrace is the API to use when capturing backtrace for
 // various tcmalloc features. It has optional emergency malloc
 // integration for occasional case where stacktrace capturing method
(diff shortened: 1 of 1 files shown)


alk:
Thanks for the patch. So as currently used this define is intended as completely internal. It not only disables sampling but also indicates that we don't have any stack trace capturing stuff built. If we want to let people turn off sampling, but keep other "non-minimal" bits, we shouldn't castrate tcmalloc::GrabBacktrace. So perhaps another define ?

alk:
Gentle ping. So what is the status here? You still want to  land something along this lines?

alk:
another gentle ping

alk:
Okay lets close then. So from what I recall, my proposal was, if anyone wants to have a define for users to disable sampling, we make it new and distinct define (not NO_TCMALLOC_SAMPLES which is for internal use).

PR #1553, opened 2024-09-01
make ThreadCache constructor/destructor private

lano1106:
1. This documents the intent that the way to create/destroy a ThreadCache object is through the static methods NewHeap()/DeleteCache()
2. It makes using the class less error-prone. The compiler will complain if some new code accidentally creates objects directly
3. This may allow some compilers to optimize code knowing that those functions are private


alk:
Applied. Thanks.

Issue #1549, opened 2024-08-28
observed performance regression


lano1106:
I was evaluating switching to the new google maintained tcmalloc project. As part of the evaluation, I did rerun the benchmark that Sanjay initially did at:
https://gperftools.github.io/gperftools/tcmalloc.html

I do not know what, but it seems like the 2024 gperftools tcmalloc is much less performant than what was there back then...

this is mostly flagrant when you compare the 8 threads 64 bytes allocation Mops.

https://github.com/google/tcmalloc/issues/260#issuecomment-2312896329

It went from 8 Mops to 2 Mops...

I do not know what but it seems like something did seriously break in between...

lano1106:
I have included jemalloc into the benchmark...

something suspicious is that all allocators appears to suffer from a lock contention except glibc... what could be the common denominator?

I have no clue on how the thread_local keyword is implemented but this could be this guy...

![mops_vs_threads_jemalloc](https://github.com/user-attachments/assets/1daf381d-4a7a-4785-b8a7-85e048aaa428)

lano1106:
I went as far as recompiling tag perftools-1.0 and the same crappy performance is found with this version so the problem is external to the project code...

I fail to see what it could be...

that would be really nice if someone could test:

```
LD_PRELOAD=libtcmalloc_minimal.so time ./ptmalloc_test1 1 1 1000000 64
LD_PRELOAD=libtcmalloc_minimal.so time ./ptmalloc_test1 8 8 1000000 64
```

to validate that he sees the same performance deterioration as soon as several threads are used.

lano1106:
yes I have compiled src/tests/ptmalloc/t-test1.c.

not with -O2 but with -O3.. beside that this is what I have done

lano1106:
my investigation has made me find a minor glitch in the project code... I'll prepare a pull request to share this finding soon...

lano1106:
I have done this:

```
diff --git a/src/thread_cache.cc b/src/thread_cache.cc
index 3ac1019..c88b7de 100644
--- a/src/thread_cache.cc
+++ b/src/thread_cache.cc
@@ -80,7 +80,8 @@ ThreadCache::ThreadCache() {
   ASSERT(Static::pageheap_lock()->IsHeld());
 
   size_ = 0;
-
+  central_fetch_cnt_ = 0;
+  central_release_cnt_ = 0;
   max_size_ = 0;
   IncreaseCacheLimitLocked();
   if (max_size_ == 0) {
[...]

PR #1554, opened 2024-09-02
introduce 2 new env vars to config ThreadCache initial state

lano1106:
new variables:
TCMALLOC_BATCH_SIZE_FROM_START:
Set the thread cache max length size to num_objects_to_move(cl) if defined

The default behavior is to set the initial max_length to 1 and increment it by 1 each time objects are fetched. Once the max_lenght reaches num_objects_to_move(cl), the max_length is now incremented by num_objects_to_move(cl) as it has been demonstrated that this particular size is in high demand.

This is done that way to minimize memory wasting. Bigger blocks are allocated only once the need for this size class has been confirmed.

If the possibility of wasting some memory is acceptable, enabling this option to eliminate the ramp-up period will:
- reduce number of accesses to the central heap
- Preserve the data locality of the objects in opposition to have the address space distributed evenly across all the caches if a particular size is requested by all threads.

This second point is not only valid when the client code access the allocated memory but also when the memory is deallocated when it is put back into a free list.

TCMALLOC_INITIAL_THREAD_CACHE_BYTES:
Set the initial cache max_size_ value.

The default behavior is to initialize the value to kStealAmount (64Kb) and increment it by kStealAmount each time a Scavenge() is triggered when a cache size reach the max_size value. This approach is the best to efficiently distribute the heap across all the caches but this imply several accesses to the central heap.

If you know the number of created threads by a process this env variable allows you to manually distribute the heap across all the threads and eliminates several central heap accesses needed to reach their stable size.

results:
--------
The after results have been obtained with the following setting:
LD_PRELOAD=/usr/lib/libtcmalloc_minimal.so TCMALLOC_RELEASE_RATE=0 TCMALLOC_TRANSFER_NUM_OBJ=1024 TCMALLOC_MAX_TOTAL_THREAD_CACHE_BYTES=100000000 TCMALLOC_INITIAL_THREAD_CACHE_BYTES=2000000 TCMALLOC_BATCH_SIZE_FROM_START=1

This appears to be an experiment that turns out to be inconclusive. With some parameters, the performance is better,
with some others the performance is worse.

It is NOT, by far, a clear-cut winner. I am not convinced myself that this should be integrated in the project.
I mostly share this experiment to stimulate and influence other experiments in that area to improve the project...
maybe with a different approach or angle, some tweaking of the idea, someone could make it work better.

mops_vs_threads_init_cache_size
mops_vs_size_init_cache_size

### DIFF
--- a/src/thread_cache.cc
+++ b/src/thread_cache.cc
@@ -70,6 +70,8 @@ volatile size_t ThreadCache::per_thread_cache_size_ = kMaxThreadCacheSize;
 
 std::atomic<size_t> ThreadCache::min_per_thread_cache_size_ = kMinThreadCacheSize;
 size_t ThreadCache::overall_thread_cache_size_ = kDefaultOverallThreadCacheSize;
+size_t ThreadCache::initial_thread_cache_size_ = kStealAmount;
+bool ThreadCache::use_batch_size_from_start_ = false;
 ssize_t ThreadCache::unclaimed_cache_space_ = kDefaultOverallThreadCacheSize;
 PageHeapAllocator<ThreadCache> threadcache_allocator;
 ThreadCache* ThreadCache::thread_heaps_ = NULL;
@@ -82,7 +84,7 @@ ThreadCache::ThreadCache() {
   size_ = 0;
 
   max_size_ = 0;
-  IncreaseCacheLimitLocked();
+  SetInitialLimitLocked();
   if (max_size_ == 0) {
     // There isn't enough memory to go around.  Just give the minimum to
     // this thread.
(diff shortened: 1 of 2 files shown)


lano1106:
this modif idea came while investigating https://github.com/gperftools/gperftools/issues/1549

alk:
Thanks for the patch. Yes, I am not convinced too. Mostly on the grounds of keeping things simpler. More tuning knobs means we have to test more. And more chances for more odd behaviors.

Patch-wise, I only want to point out that in Google C++ style and this project we always do braces even for single statement if/else things as you can see around places you changed. So for future changes (or this one if you somehow find good reasons to merge), please adjust accordingly.

lano1106:
ok. I have taken good note of the coding style adopted by the project... I try to look around what the code look like and respect as much as I can the style that I see... but I definitely drop few ones unintentionally.

I have mostly contributed to linux kernel lately which has a totally different style...

alk:
So, closing under assumption that this is exposing too much of internal details as noted above. Thanks for suggestion, though. Feel free to reopen if I misunderstood and you still want some version of this merged.

Issue #1564, opened 2024-10-02
2.16: `sampling_debug_test` fails


Lastique:
In gperftools 2.16, `sampling_debug_test` almost always fails on Kubuntu 24.04. Here is an output of me running the test under gdb, with a backtrace of the crash:

```
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
pprof cmdline: "./src/pprof" --text "/build/gperftools/gperftools/.libs/sampling_debug_test" "/tmp/sampling_test.heap.ZOprgb"
[Detaching after vfork from child process 155592]
Using local file /build/gperftools/gperftools/.libs/sampling_debug_test.
Using local file /tmp/sampling_test.heap.ZOprgb.
Adjusting heap profiles for 1-in-524288 sampling rate
Heap version 2
pprof output:
Total: 87.2 MB
    79.3  90.9%  90.9%     79.3  90.9% MallocBlock::Allocate (inline)
     7.9   9.1% 100.0%      7.9   9.1% AddressMap::Insert (inline)
     0.0   0.0% 100.0%      7.4   8.5% AddressMap::FindCluster (inline)
     0.0   0.0% 100.0%      7.9   9.1% AddressMap::Insert
     0.0   0.0% 100.0%      7.9   9.1% AddressMap::New (inline)
     0.0   0.0% 100.0%     87.2 100.0% AllocateAllocate.isra.0
     0.0   0.0% 100.0%     87.2 100.0% MallocBlock::Allocate
     0.0   0.0% 100.0%      7.9   9.1% MallocBlock::Initialize (inline)
     0.0   0.0% 100.0%     87.2 100.0% __libc_start_call_main
     0.0   0.0% 100.0%     87.2 100.0% __libc_start_main_impl
     0.0   0.0% 100.0%     87.2 100.0% _start
     0.0   0.0% 100.0%     87.2 100.0% main


AllocateAllocate regex match: 87.2 100.0% AllocateAllocate
pprof cmdline: "./src/pprof" --text "/build/gperftools/gperftools/.libs/sampling_debug_test" "/tmp/sampling_test.growth.nPumGg"
[Detaching after vfork from child process 155649]
Using local file /build/gperftools/gperftools/.libs/sampling_debug_test.
Using local file /tmp/sampling_test.growth.nPumGg.
pprof output:
Total: 103.0 MB
[...]

PR #1565, opened 2024-10-02
Avoid sampling_test failures

Lastique:
Increase the upper bound of allocated memory amount to avoid test failures on Kubuntu 24.04.

Fixes https://github.com/gperftools/gperftools/issues/1564.


alk:
Applied. thanks.

PR #1566, opened 2024-10-03
Fix gcc 10 compilation error

Lastique:
Apparently, gcc 10 doesn't support trailing return types for lambdas with an attribute.

```
src/tests/sampling_test.cc: In lambda function:
src/tests/sampling_test.cc:70:56: error: expected '{' before '->' token
   70 |   auto local_noopt = [] (void* ptr) ATTRIBUTE_NOINLINE -> void* {
      |                                                        ^~
src/tests/sampling_test.cc: In function 'void* AllocateAllocate()':
src/tests/sampling_test.cc:70:56: error: base operand of '->' has non-pointer type 'AllocateAllocate()::<lambda(void*)>'
src/tests/sampling_test.cc:70:59: error: expected unqualified-id before 'void'
   70 |   auto local_noopt = [] (void* ptr) ATTRIBUTE_NOINLINE -> void* {
      |                                                           ^~~~
```

Remove the trailing return type as it is deduced from the `noopt` call anyway.

gcc 10.2.1 is used as the system compiler in Debian 11 Bullseye.

### DIFF
--- a/src/tests/sampling_test.cc
+++ b/src/tests/sampling_test.cc
@@ -67,7 +67,7 @@ static std::string NaiveShellQuote(std::string_view arg) {
 
 extern "C" ATTRIBUTE_NOINLINE
 void* AllocateAllocate() {
-  auto local_noopt = [] (void* ptr) ATTRIBUTE_NOINLINE -> void* {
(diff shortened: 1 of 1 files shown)


alk:
Applied. Thanks.

PR #1575, opened 2024-12-13
Add CmakeList.txt install

aggresss:


alk:
Thanks for the patch, but it is not right.

First, common and stacktrace are not the libraries we intend to be installed and used by 3rd party code. They're "convenience archives" that are linked into libtcmalloc, libtcmalloc_minimal etc.

Second, what is that thing about DESTINATION set to lib/gperftools ?

alk:
This is still not right. See above.

What kind of convenience w.r.t. system integration are you referring to? (I am genuinely curious)

 Autotools-based ./configure && make works and enables all kinds of integrations. Or at least intended to.

Issue #800, opened 2016-05-03
Open-source bazel BUILD file


s-kanev:
Bazel seems to be picking up steam. And I'm fairly certain that blaze BUILD files for gperftools do exist. Any chance of open-sourcing them as an experimental second build system?

The current autotools setup makes certain build situations unnecessarily difficult (e.g. an FDO build of many targets that depend on tcmalloc, including doing FDO for tcmalloc itself), and bazel/blaze shine for such builds.

alk:
Hi Svilen. I know very little about blaze/bazel and I am very short on time to do that myself.

But if you have access yourself and can contribute this, or convince somebody else to contribute then I'll happily accept and maintain it.

siddharthab:
I found an example bazel BUILD file at https://github.com/bazelment/trunk/blob/master/third_party/gperftools/BUILD

Looks like this was tested against version 2.4, so will need some updates to run against 2.6.1. I found others but this is the one that seemed to understand your Makefile the best.

There seems to be a bazel discussion around using tcmalloc in this thread --
 https://github.com/bazelbuild/bazel/issues/818 -- so I am guessing that's another BUILD file out there.

We have also created our own BUILD file but find it difficult to maintain. Seeing that many people have tried to reverse engineer your make process in bazel, I think it is a worthwhile exercise if you could maintain it along with your Makefile.

Thank you.

jmillikin-stripe:
I don't have an internal BUILD file for gperftools / tcmalloc. Once upon a time I gave it a shot, but the Autotools config macros and compiler intrinsics were too complex to migrate within my time budget. https://github.com/gperftools/gperftools/blob/master/m4/acx_nanosleep.m4 and https://github.com/gperftools/gperftools/blob/master/m4/pc_from_ucontext.m4 were particularly challenging.

alk:
It will be fine to add bunches of preprocessor ifdefs to deal with this on common OSes.

I am fine taking contributed bazel build files, but note that I will only maintain it on best effort basis (so assume worst effort).

htuch:
@alk sounds good. I'm going to start an internal thread on this as well, as it's be ideal to see what we can do to help provide a cohesive experience for adopters of two Google technologies here (Bazel + gperftools).

rosun82:
Is there any update on this?

cas4ey:
> Is there any update on this?

+1

gorakhargosh:
+1

altonchuzhan:
+1

[...]

PR #1580, opened 2025-01-16
Add CMake install TARGETS support.

faximan:
This offers an alternative to https://github.com/gperftools/gperftools/pull/1575, which seems to have gone stale.

Why we want this: this PR allows us to depend on the built `.so` in a Bazel project using CMake through https://github.com/bazel-contrib/rules_foreign_cc. Specifically, we have successfully integrated this with

```
load("@rules_foreign_cc//foreign_cc:defs.bzl", "cmake")

filegroup(
    name = "all_srcs",
    srcs = glob(["**"]),
)

cmake(
    name = "gperftools",
    cache_entries = {
        "BUILD_TESTING": "OFF",
        "GPERFTOOLS_BUILD_CPU_PROFILER": "ON",
        "GPERFTOOLS_BUILD_DEBUGALLOC": "OFF",
        "GPERFTOOLS_BUILD_HEAP_CHECKER": "OFF",
        "GPERFTOOLS_BUILD_HEAP_PROFILER": "ON",
        "GPERFTOOLS_BUILD_MINIMAL": "OFF",
        "gperftools_enable_libunwind": "OFF",
    },
    copts = ["-Wno-error"],
    env = {
        "CMAKE_BUILD_TYPE": "Release",
        "CMAKE_BUILD_PARALLEL_LEVEL": "4",
    },
    lib_source = ":all_srcs",
    out_shared_libs = [
        "libprofiler.so",
        "libtcmalloc.so",
    ],
)

```

It is possible this is also doable with autotools/make, but we ran into problems getting Bazel/gperftools to accept properties of our Clang based C++ toolchain. The CMake integration was much more straightforward.

This might also be interesting wrt to https://github.com/gperftools/gperftools/issues/800.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -539,6 +539,7 @@ set(MINIMAL_MALLOC_SRC
   src/malloc_extension.cc)
 
 add_library(tcmalloc_minimal ${TCMALLOC_CC} ${MINIMAL_MALLOC_SRC})
+install(TARGETS tcmalloc_minimal)
 target_compile_definitions(tcmalloc_minimal PRIVATE NO_TCMALLOC_SAMPLES)
 target_link_options(tcmalloc_minimal INTERFACE ${TCMALLOC_FLAGS})
 target_link_libraries(tcmalloc_minimal PRIVATE common)
@@ -704,6 +705,7 @@ if(GPERFTOOLS_BUILD_DEBUGALLOC)
   endif()
 
   add_library(tcmalloc_minimal_debug src/debugallocation.cc ${MINIMAL_MALLOC_SRC})
+  install(TARGETS tcmalloc_minimal_debug)
   target_compile_definitions(tcmalloc_minimal_debug PRIVATE NO_TCMALLOC_SAMPLES)
   target_link_libraries(tcmalloc_minimal_debug PRIVATE symbolize low_level_alloc common)
 
@@ -776,6 +778,7 @@ if(GPERFTOOLS_BUILD_HEAP_PROFILER)
     src/malloc_backtrace.cc
     src/heap-checker-stub.cc)
   add_library(tcmalloc ${TCMALLOC_CC} ${FULL_MALLOC_SRC})
+  install(TARGETS tcmalloc)
   target_compile_definitions(tcmalloc PRIVATE ${EMERGENCY_MALLOC_DEFINE})
   target_link_libraries(tcmalloc PRIVATE stacktrace low_level_alloc common)
   target_link_options(tcmalloc INTERFACE ${TCMALLOC_FLAGS})
@@ -831,6 +834,7 @@ endif()
 if(GPERFTOOLS_BUILD_DEBUGALLOC)
   if(GPERFTOOLS_BUILD_HEAP_PROFILER)
     add_library(tcmalloc_debug src/debugallocation.cc ${FULL_MALLOC_SRC})
+    install(TARGETS tcmalloc_debug)
     target_compile_definitions(tcmalloc_debug PRIVATE ${EMERGENCY_MALLOC_DEFINE})
     target_link_libraries(tcmalloc_debug PRIVATE symbolize low_level_alloc stacktrace common)
 
@@ -857,10 +861,11 @@ endif()
 ### ------- CPU profiler
 if(GPERFTOOLS_BUILD_CPU_PROFILER)
 
-add_library(profiler
-    src/profiler.cc
-    src/profile-handler.cc
-    src/profiledata.cc)
+  add_library(profiler
+      src/profiler.cc
+      src/profile-handler.cc
+      src/profiledata.cc)
+  install(TARGETS profiler)
   target_link_libraries(profiler PRIVATE stacktrace common)
 
   if(BUILD_TESTING)


alk:
Thanks for the patch. But this is also not good enough. I see only .so files installed. And looks like "proper" install support for cmake would involve some degree of .cmake file or(and) .pc file, to integrate "properly".

But I am not sure if I want to have "it" merged (and being maintained by me), given how extremely messy this entire packaging/installing thing is in cmake worlds.

mering:
> Thanks for the patch. But this is also not good enough. I see only .so files installed. And looks like "proper" install support for cmake would involve some degree of .cmake file or(and) .pc file, to integrate "properly".

I think this PR also ships improvements in itself:
- `FindGperftools.cmake` files can also be maintained downstream
- This allows distros using the CMake workflows for packaging (usually making use of `DESTIR` and `PREFIX` variables)
- This unblocks Bazel support via rules_foreign_cc

So I would vote to merge this as is and add `.cmake` or `.pc` files in a follow up if you really want to maintain this. As I understand, the same would be useful also for the Autotools workflow, so I assume this request is independent from this PR.

alk:
Whatever changes you need to unblock whatever you need can also be maintained downsteam. Autotools workflow already supports everything people should need in 100% standard and idiomatic way (well, "standard" in gnu/autotools sense). This includes shipping .pc file.

Please look at this from my perspective:

a) If I merge this I'll be bugged about this ~semi-constantly. And this is clearly not close to 100% usable. I'd rather be bugged about install not being present, than having broken cmake install.

b) If I merge this I implicitly agree to get this into more usable shape and maintain it. And I am still unsure about fate of entire cmake-for-gperftools thingy. Because maintaining cmake parts is extra work of unclear utility.

c) I could use help defining what "done" means here. I inspected some other packages heavily using cmake (e.g. gemini recommended me to look at fmtlib) and I don't like how it looks. I am unsure what combination of cpack, .cmake, .pc makes sense to do. And on top of that, it looks like some people would like to be able to "vendor" gperftools and integrate it's build system into theirs (without intermediate install and "find gperftools" step).

In short, I am ready to merge something that works for some reasonable definition of works (see above; but at minimum if someone install gperftools this way it should be 100% usable) and that isn't too complicated or un-idiomatic cmake-wise.

alk:
Hi and thanks for the response. There is still certain quality bar for contributions. This thing doesn't even get headers installed. I don't consider such state as usable. Once it installs headers, then it'll pass minimum quality bar for improvement.

mering:
> Hi and thanks for the response. There is still certain quality bar for contributions. This thing doesn't even get headers installed. I don't consider such state as usable. Once it installs headers, then it'll pass minimum quality bar for improvement.

For which libraries do you need headers? For `libprofiler.so` and `libtcmalloc.so` it is enough to link against the libraries to produce the desired output as described in your README.

alk:
okay, easier for me to accept this then argue. Indeed, .so libraries alone can be useful occasionally. But I amended it with a change that made install targets disabled by default. See here: https://github.com/gperftools/gperftools/commit/0576523a6028626a7760dabf591c6713d6f69d73

alk:
ah, btw, if this is not a secret. That blaze example above is it for some sort of open-source project or something closed? Asking in case I could help make that integration better.

mering:
@alk Thanks for merging!

Once you tag a new release, we plan to upload the BUILD file to [Bazel Central Registry](https://registry.bazel.build/) which is basically a Bazel package manager such that it can easily be used in any Bazel workspace.

---

Our documentation for using it (all kudos to @faximan as the author):

1. Link the profiler into your binary by specifying `linkopts` and `deps`:

   ```diff
   cc_test(
       name = "foo_test",
       srcs = ["foo_test.cc"],
   +   linkopts = [
   +       "-lprofiler",  # For CPU profiling
   +       "-ltcmalloc",  # For heap profiling
   +   ],
       deps = [
           ":foo",
   +       "@gperftools",
       ],
   )
   ```

1. Run the binary with an environment variable set:

   ```bash
   # For CPU profiling
   CPUPROFILE=/tmp/cpu_profile bazel run --config intrinsic :foo_test

   # For Heap profiling
   HEAPPROFILE=/tmp/heap_profile bazel run --config intrinsic :foo_test
   ```

   Note: you need `bazel run`, also for `cc_test`, to allow a profile to be written to disk.

alk:
I am glad I asked. Doing bazel registry integration via our cmake integration is IMHO too un-excellent. And hardcoding stuff like Release or number of jobs is just one of imperfections.

We can do better. I was actually planning for this year to improve our integration with those "c++ package managers". So perhaps you can help me understand how to do it better for your blaze thingy:

a) I see that both "abseil" tcmalloc and mimalloc in that registry are far from most recent. How is that registry updated? Can we somehow automate it? Who is generally responsible for that?

b) If I provide less unexcellent bazel build how much extra work is to do registry integration?

mering:
Thanks for your willingness to improve things further.

> I am glad I asked. Doing bazel registry integration via our cmake integration is IMHO too un-excellent. And hardcoding stuff like Release or number of jobs is just one of imperfections.

Perfect is the enemy of good 😉  The BCR ecosystem is just getting started and having something working is still better than having nothing. Things can be improved in follow-ups. Especially for dependencies which only change when explicitly bumping a version, they are used from cache most of the time anyways, so a little slower build than possible doesn't do a lot of harm.

> We can do better. I was actually planning for this year to improve our integration with those "c++ package managers". So perhaps you can help me understand how to do it better for your blaze thingy:
> 
> a) I see that both "abseil" tcmalloc and mimalloc in that registry are far from most recent. How is that registry updated? Can we somehow automate it? Who is generally responsible for that?

Best case, the upstream repo has Bazel support and publishes to BCR itself. There is some automation available (for example used by Protobuf) and some more sophisticated automation is being developed.
If the upstream project doesn't upload themselves then it is up to the community to do the upload (similar to Linux distros). So feel free to send a PR to upload new versions of dependencies you care about.

> b) If I provide less unexcellent bazel build how much extra work is to do registry integration?

I think the registry upload is not the hardest problem. Maintaining proper Bazel BUILD files for your repo is. Given that you have not been very keen on maintaining CMake support, I am wondering how much you like maintaining one more build system natively.

alk:
Okay, I could use help reviewing and possibly fixing basic bazel support here: https://github.com/alk/gperftools/compare/master...wip-bazel

MSVC detecting is clearly not working. And windows detecting is semi-borked as well. And could use help making sure only tcmalloc_minimal and relevant programs and tests are enabled on windows.

mering:
@alk Sorry for the late answer as I was OOO last week. From looking at the commits in master it seems like you figured it out. Unfortunately, I don't have experience with using Bazel on non Linux platforms.

PR #1594, opened 2025-05-14
Adding 'sb' instruction to spin_delay() for ARM v8.5 onward

salvatoredipietro:
We would like to propose this optimization for ARM architecture that, at runtime, it switches to SB instruction if supported by the system. SB (Speculation Barrier) is a modern barrier which is available from armv8.5a. It achieves the same result as issuing ISB, but instead of flushing the CPU it does so by serializing older instructions to be non-speculative before it completes. This is less disruptive than an "isb" to high performance CPUs.

We already saw positive improvements on MySQL server (https://github.com/mysql/mysql-server/pull/611) and Folly (https://github.com/facebook/folly/pull/2390) implementations.

### DIFF
--- a/src/base/spinlock.cc
+++ b/src/base/spinlock.cc
@@ -37,6 +37,10 @@
 #include "base/spinlock_internal.h"
 #include "base/sysinfo.h"   /* for GetSystemCPUsCount() */
 
+#if defined(__GNUC__) && defined(__aarch64__)
+#include <sys/auxv.h>
+#endif // end __aarch64__
+
 // NOTE on the Lock-state values:
 //
 // kSpinLockFree represents the unlocked state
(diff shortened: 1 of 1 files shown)


alk:
Thanks for the patch. I trust you on the benefits of SB instructions usage, but actual details need a bit more work.

So sys/auxv.h header seems to be linux-only (not gcc or gcc-compatible compiler-specific as in your patch). As is getauxval. And I think you want to ifdef around availability of HWCAP_SB. Or perhaps you can define it if it is missing in people's specific glibc.

Do you also want to try to do it for OSX? Their cpus are seemingly quite modern instruction-set-wise, so we "just" need to do right detection there.

Similarly some windows machines are actually running on modern ARMs this days so maybe we can make it work there too (mingw and msvc if possible).

salvatoredipietro:
Thanks for the feedbacks.

- Regarding the sys/auxv.h wrap, I was copying the same condition where `isb` is currently defined (`#elif defined(__GNUC__) && defined(__aarch64__)`). Not sure about all the implications of change these limits in the project. Let me know how you would like to set them
- HWCAP_SB definition, yes I can add it, no problem.
- Regarding MacOS and Windows: I do not have the possibility to test it. My Mac is on ARMv8.4 which does not support it. If you have ability to do it, I will be happy to integrate them
- __builtin_expect: I will change it in the follow up commit to PREDICT_TRUE macro.

alk:
ISB is used unconditionally. Ifdef is then to detect that compiler understands gnu-style inline assembly. getauxval is linux-specific and requires carerp in detection and usage

salvatoredipietro:
I see, thanks for the clarification. Would like that be enough?

alk:
Couple more things. Why do you append \n at the end of all asm ? Please don't.

if (getauxval(..) & HWCAP_SB) while works, isn't really good enough. Either add extra braces, or have explicit != 0 comparison. I believe GCC should be giving you warning there.

In this project we also prefer not to have "naked" statements in if/else branches. Brace them with {}.

Also I think it'll be all much nicer if you place detection logic into SpinLock_InitHelper thingy, where we already have some initialization/detection logic. Then you can simply have both instructions in code (even on OSX/Windows) but SB will be switched "on" when detection enabled it.

Also what is the most simple way for me to ensure that your .instr thingy is indeed encoding of SB (and not something else, possibly nefarious) ?

salvatoredipietro:
Addressed the above feedback in the latest commit.

> Also what is the most simple way for me to ensure that your .instr thingy is indeed encoding of SB (and not something else, possibly nefarious) ?

I have used to search for it in the output, for example like this:
```
$ objdump -d libcommon.a | grep  sb 
  1c:   d50330ff        sb
  40:   d5033fdf        isb
```

alk:
osx failure seems unrelated. I'll handle it separately. On first quick sight it looks fine. I'll have another look some time soon. Thanks a lot.

PR #1604, opened 2025-07-26
perferring normal spans to returned spans in PageHeap::AllocLarge

### LINKED ISSUE #1574: ReleaseAtLeastNPages always madvise pages with smallest addresses, causing pointless decommit for large spans
In the `PageHeap::ReleaseAtLeastNPages` function, we decommit spans from `large_normal_.begin()`, which have smallest length and address, and then move them to `large_returned_`

However, in `PageHeap::AllocLarge`, spans in `large_returned_` with smaller start address take priority over those in `large_normal_`.  As a result, it’s likely that the span we select is the same one we just decommitted in the previous `PageHeap::ReleaseAtLeastNPages` call.

This behavior can lead to slow decommit as well as performance degradation.

So is there any specific reason for decommitting spans from `large_normal_.begin()` instead of `large_normal_.rbegin()` in `PageHeap::ReleaseAtLeastNPages` ?

code from `PageHeap::ReleaseAtLeastNPages` :
https://github.com/gperftools/gperftools/blob/ab499855dc9477834ca88cae585befb8fa15fe95/src/page_heap.cc#L612-L622

code from `PageHeap::AllocLarge`:
https://github.com/gperftools/gperftools/blob/ab499855dc9477834ca88cae585befb8fa15fe95/src/page_heap.cc#L289-L298

JackieCui00:
Fixes #1574

### DIFF
--- a/src/page_heap.cc
+++ b/src/page_heap.cc
@@ -292,8 +292,7 @@ Span* PageHeap::AllocLarge(Length n) {
     Span *c = place->span;
     ASSERT(c->location == Span::ON_RETURNED_FREELIST);
     if (best_normal == nullptr
-        || c->length < best->length
-        || (c->length == best->length && c->start < best->start))
+        || c->length < best->length)
       best = place->span;
   }
 
--- a/src/tests/page_heap_test.cc
+++ b/src/tests/page_heap_test.cc
@@ -11,6 +11,7 @@
 
 #include <limits>
 #include <memory>
+#include <vector>
 
 #include "page_heap.h"
 
@@ -80,6 +81,51 @@ TEST(PageHeapTest, Stats) {
   // Delete span 's1'
   ph->Delete(s1);
   CheckStats(ph.get(), 256, 128, 128);
+}
+
+TEST(PageHeapTest, Decommit) {
+  if (!HaveSystemRelease()) {
+      return;
+  }
+
+  std::unique_ptr<tcmalloc::PageHeap> ph(new tcmalloc::PageHeap());
+
+  constexpr size_t kNumPtrs = 10;
+  constexpr size_t kBigAllocPages = kMaxPages * 2;
+
+  std::vector<tcmalloc::Span*> used_spans;
+  std::vector<tcmalloc::Span*> free_spans;
+  for (size_t i = 0; i < kNumPtrs; ++i) {
+    // interleave free_spans and used_spans to prevent free_spans from coalescing
+    free_spans.push_back(ph->New(kBigAllocPages));
+    used_spans.push_back(ph->New(kBigAllocPages));
+  }
+
+  for (auto span : free_spans) {
+    ph->Delete(span);
+  }
+
+  for (size_t i = 0; i < 2 * kNumPtrs; ++i) {
+    {
+      SpinLockHolder l(ph->pageheap_lock());
+      EXPECT_EQ(kBigAllocPages, ph->ReleaseAtLeastNPages(1));
+    }
+
+    tcmalloc::Span* new_span = ph->New(kBigAllocPages);
+    ph->Delete(new_span);
+  }
+
+  {
+    SpinLockHolder l(ph->pageheap_lock());
+    EXPECT_EQ(kBigAllocPages, ph->ReleaseAtLeastNPages(1));
+  }
+  for (auto span : free_spans) {
+    EXPECT_EQ(span->location, tcmalloc::Span::ON_RETURNED_FREELIST);
+  }
+
+  for (auto span : used_spans) {
+    ph->Delete(span);
+  }
 }
 
 // The number of kMaxPages-sized Spans we will allocate and free during the
--- a/src/tests/tcmalloc_unittest.cc
+++ b/src/tests/tcmalloc_unittest.cc
@@ -78,6 +78,7 @@
 #endif
 
 #include <algorithm>
+#include <array>
 #include <functional>
 #include <mutex>
 #include <new>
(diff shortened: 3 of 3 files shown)


alk:
Thanks for the patch. The code looks fine (there are couple minor things I'd note regarding tests but those really are minor code style/etc).

Real issue I have is how do we evaluate this heuristic. In principle, this entire thing about returning to OS, about having distinct returned and non-returned spans, breaking hugepages etc, is quite messy already. Best way I know how to deal with it (and we did this largely with some of the more important services at Google) is to disable this thing entirely. Yes, "abseil" tcmalloc now has very fancy hugepage-aware heap (and I actually code reviewed most commits there), but I am not entirely buying it either.

If only we could somehow find some workloads, even mostly/largely synthetic, that could actually exercise this stuff. So we know what to actually optimize....

BTW, how did you find this issue? Was it just studying the code, or are you running something that actually exercises the issue you're trying to fix?

JackieCui00:
Yes, I discovered this issue during online troubleshooting.

Our service is an online storage system, which frequently calls `AllocLarge` for data buffers. After a burst of IO, we often see the heap/RSS grows, sometimes by more than 10 GiB.

Actually, our application periodically calls `ReleaseToSystem` when RSS exceeds a certain threshold,  and to avoid performance jitters, we only release a small amount of memory at a time (for example, 128 MiB per call).

However, we noticed that even after running `ReleaseToSystem` repeatedly over a long period, the RSS would not decrease as much as expected. This led us to investigate further and eventually identify the problem of `ReleaseAtLeastNPages`.

As shown in the UT of this patch, with the current implementation, if the rate of decommit via `ReleaseToSystem` cannot keep up with the allocation rate, the RSS will not decrease.

JackieCui00:
I completely agree with you that evaluating the solution is indeed the challenging part. However, I believe there is no doubt that the current implementation needs improvement.

For the scenario of our online service, I think both the LRU solution and Go Runtime approach (always returning the span with largest address) would be effective, as the program typically performs `AllocLarge` for only a few categories of sizes, or even just one specific size.
However,  the LRU solution is much more easier to implement, especially since the "not-large" spans are already managed using the LRU strategy.

JackieCui00:
> Also in general, why having 10 gigs of unreleased RSS (as point out above) is the problem? Workloads usually size for the peak memory usage anyways. So unused at some point, but ready to be quickly allocated memory isn't necessarily bad thing.

Sorry for not being clear about what I meant by "a burst of IO".  In our system, such bursts are typically triggered by unexpected events, like network retransmissions or disk failures. These incidents cause IO latencies to increase significantly, resulting in a large number of IOs pending in our system, which in turn leads to a spike in RSS.

For us, this scenario is considered abnormal and needs to be recovered from as quickly as possible. That’s why it's important for us to return the extra RSS consumed during these bursts back to the OS once the situation returns to normal.

JackieCui00:
> Also if you somehow care about releasing unused memory off peak (and I still don't quite understand why), have you considered "simply" doing aggressive decommit mode?

Actually, we have deliberately disabled "aggressive decommit mode" in our system. 

Since "aggressive decommit mode" can only be enabled or disabled globally, enabling it means that the critical IO path may call `madvise` under normal circumstances, which maybe increases IO latency by tens or even hundreds of microseconds.
We have observed this performance impact in practice, and chose to disable "aggressive decommit mode" during startup.

Besides, even we call `ReleaseToSystem` when RSS becomes too large, we call it in a controlled manner: we only return a small chunk of memory (such as 128MiB) at a time, and do so in a background thread.

alk:
Thanks for good explanations. Thinking about this some, I think the core issue is that AllocLarge always prefers "better" returned span to the normal span. This is not how we handle non-large span allocations.

I suspect this logic made sense before I have added "return ~everything before we are forced to grow heap" logic. This:https://github.com/gperftools/gperftools/blob/338e906eaec5cd136f392fe81d533935aafb0777/src/page_heap.cc#L188 . I.e. if we always prefer good enough normal span to returned span, we'll sorta encourage set of returned spans to grow and possibly fragment. Which will/could occasionally cause total heap size to grow. So avoiding this they probably choose to always prefer better fit.

Of course this entire separation of normal/returned spans is quite imperfect. So if we could find much more principled way to handle this, I'd be willing to try it.

How about we actually try to make LargeAlloc always prefer good enough normal span. The logic is, we know that touching returned memory is kinda expensive, so we should try to avoid this. This looks a bit larger and thus riskier change, but it also seems like better one.

We can also consider making old/new behavior probabilistic. I.e. most of the time we prefer to avoid touching returned spans, but once in a while we do take returned span if it is better fit (and we can define better to be much stricter than plain-old address-ordered-best-fit).

What do you think?

JackieCui00:
I've updated the patch based on the idea of preferring normal spans in `PageHeap::AllocLarge`, and it has passed the unit tests as expected.
I would greatly appreciate your feedback on this.

alk:
This looks good. Lets get this in. Looks like you're lacking include <array>.

I am now taking a closer look at your testing code. One minor notice there is we try to follow Google C++ style. So NUM_PTRs and it's neighbor should be kNumPtrs. Also perhaps make them constexpr.

alk:
The rest of the tests code looks fine. If you can avoid auto new_span in the test and explicitly declare a new_span variable as Span* that would be even better (this is also as per Google C++ Style thing).

JackieCui00:
Yes, I have updated the patch to address the style issues and fixed the unit test failures.

Please let me know if you have any further suggestions or feedback.

alk:
This looks good. I will get this merged sometime very soon. I was about to merge, but there is actually somewhat subtle flakiness in the ReleaseToSystem test, which becomes visible after your changes. I'll get it fixed and then cherry-pick your code. Thanks, again, for your contribution.

alk:
Merged. Turns out bigger source of flakiness was in your change. We need to capture starting_bytes _after_ malloc-ing big chunks, since those chunks may come out of previously returned memory. I fixed this (with signoff and explicit comment in the commit message). Thanks, again.

PR #1606, opened 2025-08-07
stacktrace_libgcc-inl.h: fix build for some ARM toolchains

graysky2:
On OpenWrt when bulding for arm_cortex-a15_neon-vpfv4 or arm_cortex-a9_vfpv3-d16, the build errors out due to an undefined symbol _URC_NORMAL_STOP. This usually happens when the required header (typically <unwind.h>) is not included, or the toolchain's unwind implementation does not provide this definition.

Error:
src/stacktrace_libgcc-inl.h: In function '_Unwind_Reason_Code libgcc_backtrace_helper(_Unwind_Context*, void*)': src/stacktrace_libgcc-inl.h:69:12: error: '_URC_NORMAL_STOP' was not declared in this scope
   69 |     return _URC_NORMAL_STOP;
      |

A work-around is to add definition.


graysky2:
Yes, the project offers prebuilt tool chains for all supported hardware. I’m on my mobile now too much of a pain to browse, but I will share a link with you if you don’t find one on your own later today.

graysky2:
@alk You can download and inspect the toolchains here: https://downloads.openwrt.org/snapshots/targets/

For example, for bcm53xx: https://downloads.openwrt.org/snapshots/targets/bcm53xx/generic/openwrt-toolchain-bcm53xx-generic_gcc-14.3.0_musl_eabi.Linux-x86_64.tar.zst

alk:
Thanks. Indeed, turns out arm eabi implementation of unwind.h doesn't have URC_NORMAL_STOP, but URC_END_OF_STACK is supported since at least 2005. That is long enough. I also see that libbacktrace used END_OF_STACK. So gives me piece of mind.

alk:
My only ask is to fix commit message. It says youre adding definition, and what you're doing instead is using URC_END_OF_STACK.

Also can you wrap long lines in the commit message to be no longer than 70 chars?

graysky2:
you bet, done

PR #1621, opened 2025-11-17
Recognize SPARCv9 cache line length

alarixnia:


alk:
Merged. Thanks. Please, consider doing followup with the constructive interference size as discussed.

PR #1622, opened 2025-11-19
update BUILD.bazel and MODULE.bazel to load rules_cc

mmorel-35:
#### Description

Update BUILD.bazel and MODULE.bazel to load rules_cc

This allows to be compatible with Bazel 9 constraints


alk:
Thanks. Submitted with slight modification. We use case-sensitive sorting of .gitignore-s. So I resorted and updated commit message to indicate that.

PR #1624, opened 2025-12-04
Add BCR workflow to publish gperftools on tag creation

mmorel-35:
#### Description

This workflow will help publish to BCR after a tag is published.

Based on https://github.com/bazelbuild/bazel-central-registry/tree/main/modules/gperftools


mmorel-35:
@mering , 

we are providing the publish to BCR workflow here. I don't know if you mind that the maintainer of gperftools becomes the only one declared as MODULE maintainer in BCR ?

PR #1629, opened 2026-01-28
Enable Windows builds for tcmalloc with platform-specific BCR presubmit and comprehensive Bazel CI/CD

mmorel-35:


alk:
I apologize for the delay. So there is still way too much going on in this patch. .bazelrc stuff is wrong because it is only for the bazel root project. And should not be necessary if we do the right thing (macos stuff is exception because their toolchain is borked). We're also not going to enable :tcmalloc on windows just yet because it isn't there with other build tools. What I like most about this is we're getting CI for the bazel bits. Can you extract just that part, so we can integrate it and iterate with the rest?

PR #1635, opened 2026-03-18
GNU/Hurd support

ovalenti:
Hello !

I am making use of gperftool under GNU/Hurd, and these are the few changes I had to make to support the system. Would you consider merging them, please ?

Details:

- `PATH_MAX` is not part of posix, and Hurd does not define as there is no limit in the path size. I noticed that there is already a workaround, I extended this to other places where PATH_MAX is used.
- The Apple platforms are based on MACH, Hurd as well. So I refined the MACH checks to only `__APPLE__`

### DIFF
--- a/src/base/basictypes.h
+++ b/src/base/basictypes.h
@@ -196,4 +196,12 @@ union MemoryAligner {
 #define ALWAYS_INLINE inline
 #endif
 
+#ifndef PATH_MAX
+#ifdef MAXPATHLEN
+#define PATH_MAX MAXPATHLEN
+#else
+#define PATH_MAX 4096
+#endif
+#endif
+
 #endif  // _BASICTYPES_H_
--- a/src/base/for_each_line.h
+++ b/src/base/for_each_line.h
@@ -38,6 +38,14 @@
 #include "base/function_ref.h"
 #include "base/logging.h"
 
+#ifndef        PATH_MAX
+#ifdef MAXPATHLEN
+#define        PATH_MAX        MAXPATHLEN
+#else
+#define        PATH_MAX        4096         // seems conservative for max filename len!
+#endif
+#endif
+
 namespace tcmalloc {
 
 // Using buffer of the given size, reads some contents, find lines
--- a/src/base/for_each_line.h
+++ b/src/base/for_each_line.h
@@ -38,11 +38,11 @@
 #include "base/function_ref.h"
 #include "base/logging.h"
 
-#ifndef        PATH_MAX
+#ifndef PATH_MAX
 #ifdef MAXPATHLEN
-#define        PATH_MAX        MAXPATHLEN
+#define PATH_MAX MAXPATHLEN
 #else
-#define        PATH_MAX        4096         // seems conservative for max filename len!
+#define PATH_MAX 4096  // seems conservative for max filename len!
 #endif
 #endif
 
--- a/src/base/for_each_line.h
+++ b/src/base/for_each_line.h
@@ -38,14 +38,6 @@
 #include "base/function_ref.h"
 #include "base/logging.h"
 
-#ifndef PATH_MAX
-#ifdef MAXPATHLEN
-#define PATH_MAX MAXPATHLEN
-#else
-#define PATH_MAX 4096  // seems conservative for max filename len!
-#endif
-#endif
-
 namespace tcmalloc {
 
 // Using buffer of the given size, reads some contents, find lines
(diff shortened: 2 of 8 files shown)


alk:
Very nice. Thanks. There are some duplicates of the PATH_MAX business. Ideally we'd do something about it.

Also on my system man 7posix limits.h does mention PATH_MAX, so it is in POSIX.

But the biggest question is if Hurd is indeed a system with no upper limit on paths, then perhaps our code needs adjustment for that somehow ?

ovalenti:
Sorry, I was imprecise, it looks like PATH_MAX may be defined, but [does not have to be](https://www.gnu.org/software/hurd/community/gsoc/project_ideas/maxpath.html).

https://pubs.opengroup.org/onlinepubs/009695399/basedefs/limits.h.html

I can propose to replace all static path allocations with dynamic memory, but this is a much larger change.

> There are some duplicates of the PATH_MAX business

Do you have a strategy to suggest ? A common header file maybe ?

alk:
So straightforward dynamic memory allocation is out of question, most likely (would have to double check). A number of those places are likely deep enough in the guts of  malloc or early enough that using malloc/free or new/delete is not possible or safe.

For the "where to consolidate", it could be a new header, but I think we already have "a bunch of misc/random defines" header which is base/basictypes.h.

So I don't know if you have appetite for deeper look at this. If you don't have it, then, yes, lets hardcode something decent. Maybe even have a distinct name for it, say, TCM_PATH_MAX or something. With the comment that some OS (e.g. Hurd) have unlimited path length but we're not yet ready to handle them.

If you do have some appetite for the deeper work (maybe with the help of some powerful LLM advice; they are remarkable when it comes to grasping foreign code base) then start by checking each of uses of PATH_MAX and see how deeply/early they are. Feel free to come back with questions or suggestions.

ovalenti:
Hi @alk,

I am afraid I don't have the energy to rework entirely path management in gperftool. I followed your alternative proposal, and moved the workaround to a common place (and removed a pre-existing one).

> Maybe even have a distinct name for it, say, TCM_PATH_MAX or something.

I did not understand this suggestion.

ovalenti:
I realized that making gperftool work on this system would require much more work than I initially thought.

Thank you for your help and sorry for the noise.

alk:
No problems at all. Can you please elaborate about the changes necessary? I am mildly interested in supporting GNU/Hurd myself. So having some initial background, whatever you can tell me, would be very useful.

ovalenti:
At some point, I thought that the profiling tool was making use of the debugging facility provided by the operating system, like GDB, which would require a substantial work. Then realized that it is using PROF timers, which should be hopefully fine, and considered that I quit prematurely on this PR.

But then I found that some tests are not passing (at least on Debian hurd Sid x86_64 up) : segmentation faults in : `page_heap_test` and `check_address_test`

```
$ ./page_heap_test 
Running main() from vendor/googletest/googletest/src/gtest_main.cc
[==========] Running 3 tests from 1 test suite.
[----------] Global test environment set-up.
[----------] 3 tests from PageHeapTest
[ RUN      ] PageHeapTest.Stats
[       OK ] PageHeapTest.Stats (6 ms)
[ RUN      ] PageHeapTest.Decommit
[       OK ] PageHeapTest.Decommit (0 ms)
[ RUN      ] PageHeapTest.Limit
src/tests/page_heap_test.cc:198: Failure
Expected: (spans[i]) != (nullptr), actual: NULL vs (nullptr)

src/tests/page_heap_test.cc:198: Failure
Expected: (spans[i]) != (nullptr), actual: NULL vs (nullptr)

src/tests/page_heap_test.cc:198: Failure
Expected: (spans[i]) != (nullptr), actual: NULL vs (nullptr)

src/tests/page_heap_test.cc:198: Failure
Expected: (spans[i]) != (nullptr), actual: NULL vs (nullptr)

src/tests/page_heap_test.cc:198: Failure
Expected: (spans[i]) != (nullptr), actual: NULL vs (nullptr)

src/tests/page_heap_test.cc:198: Failure
Expected: (spans[i]) != (nullptr), actual: NULL vs (nullptr)

src/tests/page_heap_test.cc:198: Failure
Expected: (spans[i]) != (nullptr), actual: NULL vs (nullptr)

src/tests/page_heap_test.cc:198: Failure
Expected: (spans[i]) != (nullptr), actual: NULL vs (nullptr)

Segmentation fault         ./page_heap_test
```

```
$ ./check_address_test
Running main() from vendor/googletest/googletest/src/gtest_main.cc
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from CheckAddressTest
[ RUN      ] CheckAddressTest.MainAccess
Segmentation fault         ./check_address_test
```

PR #1640, opened 2026-07-23
Don't pick up vendored googletest from bazel

hzeller:
In bazel, we use the googletest from MODULE.bazel, so tell bazel to not recurse into the vendored directory.

This allows to `bazel build ...` without issue.


alk:
Thanks!

PR #1642, opened 2026-07-23
Add `:gtest` dependencies to tests using the headers.

hzeller:
Thee tests linked against `:gtest_main` that only indirectly provided these headers; but in the spirit of `dwyu` (Depend on What You Use), we should also directly depend on the gtest lib.

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -1,8 +1,8 @@
 # NOTE: as of this writing Bazel support is highly experimental. It is
 # also not entirely complete. It lacks most tests, for example.
 
-load("@rules_cc//cc:defs.bzl", "cc_binary", "cc_library", "cc_test")
 load("@bazel_skylib//lib:selects.bzl", "selects")
+load("@rules_cc//cc:defs.bzl", "cc_binary", "cc_library", "cc_test")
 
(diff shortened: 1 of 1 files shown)


alk:
Hi. Thanks for the patch. But I don't understand the reasoning. Can you elaborate why? Isn't gtest_main enough?

hzeller:
`:gtest_main` is enough in the sense that then also depends on `:gtest`, so is transitively providing the headers that originally only `:gtest` provides (gtest_main does not have any `hdrs = []` declared).

Depending on something that happens to depend on the library that actually provides the header files we're using is of course fragile, so being explicit and depending on all the libraries that directly provide the headers that we are using  makes things more robust. So this is why I added the `:gtest` library here as we use the headers.

For header files, there is the concept of `IWYU` (Include What You Use), essentially always include all the headers that you need to provide all the symbols, not depending on headers that then happen to include other headers providing the symbols then.
For dependency management, one could think of it as `DWYU` (Depend on What You Use).

(I am using the [`bant`](https://github.com/hzeller/bant) tool to automatically find the missing dependencies  (disclaimer: I wrote that tool :) ))

